PR #4, opened 2019-02-13
add initial jasync support

oshai:


jasync-sql/src/main/java/io/micronaut/configuration/jasync/JasyncPostgreSQLClientFactory.java
```diff
@@ -0,0 +1,54 @@
...
+        this.jasyncPoolConfiguration = jasyncPoolConfiguration;
+    }
+
+    /**
+     * Create a connection pool to the database configured with the {@link JasyncPoolConfiguration}.
+     * @return client A pool of connections.
+     */
+    @Singleton
+    @Bean(preDestroy = "disconnect")
+    public Connection client() {
```
graemerocher: Just checking, this bean is for the life of the whole app right? Not per request?

jasync-sql/src/main/java/io/micronaut/configuration/jasync/condition/package-info.java
```diff
@@ -21,4 +21,4 @@
  * @since 1.0
  */
 
-package io.micronaut.configuration.postgres.reactive.condition;
+package io.micronaut.configuration.jasync.condition;
```
graemerocher: Seems that some packages were renamed? These probably shouldn't be part of this change

oshai:
@graemerocher I created initial support. couple of things I am not sure about:
JasyncPostgreSQLClientFactory is only for postgres, I am not sure if/how it is possible to create another one for mysql? The only difference is that PostgreSQLConnectionBuilder should be replaced with MySQLConnectionBuilder.

In JaysncHealthIndicator I use a future and then I want to convert it to a publisher, not sure I did it in a preferred way.

Tests are not passing yet, I am not sure how it works exactly.

You're help is appreciated.

jasync-sql/src/main/java/io/micronaut/configuration/jasync/health/package-info.java
```diff
@@ -20,5 +20,5 @@
  * @author puneetbehl
  * @since 1.0
  */
-package io.micronaut.configuration.postgres.reactive.health;
+package io.micronaut.configuration.jasync.health;
```
graemerocher: Renamed again?

jasync-sql/src/main/java/io/micronaut/configuration/jasync/health/JaysncHealthIndicator.java
```diff
@@ -0,0 +1,69 @@
...
+    public JaysncHealthIndicator(Connection client) {
+        this.client = client;
+    }
+
+    @Override
+    public Publisher<HealthResult> getResult() {
+        return s -> client.sendQuery(QUERY)
+                .whenComplete((result, error) -> {
+                    if (error != null) {
+                        s.onError(error);
```
graemerocher: In this case instead of emitting the error you probably want to emit a HealthResult of status `DOWN`. See the builder API

graemerocher:
There are some failing tests as well...

graemerocher:
I will look into this test failures and get back to you. For the factory part you would want some kind of condition I think you want to make both mysql and postgres dependencies `compileOnly` and then on the factory add 

```
@Requires(classes= PostgreSQLConnectionBuilder.class)
```

And then have another factory that requires the MySQL one.

jasync-sql/src/main/java/io/micronaut/configuration/jasync/JasyncPostgreSQLClientFactory.java
oshai: yes, it is a connection pool that should live as long as the app lives

jasync-sql/src/main/java/io/micronaut/configuration/jasync/condition/package-info.java
oshai: I just renamed packages that are copied for postgresql reactive. I think it should be fine, isn't it?

jasync-sql/src/main/java/io/micronaut/configuration/jasync/health/JaysncHealthIndicator.java
oshai: ok, will change that.

oshai:
I fixed the comments. regarding tests, I run them in trace and saw the following message:
`Resolved bean candidates [] for type: class io.micronaut.configuration.jasync.JasyncPoolConfiguration`

It might be related to error somehow.

graemerocher:
Tests pass with this diff https://gist.github.com/graemerocher/a130577676d8f2c6344dd16a8d9d591a

If you apply it then it will probably be ready to merge...

graemerocher:
Only other thing required will probably be some documentation. For example https://github.com/micronaut-projects/micronaut-sql/blob/master/src/main/docs/guide/pgclient.adoc

oshai:
I will add docs as well in few minutes.

oshai:
I added docs as well.

src/main/docs/guide/jasync.adoc
```diff
@@ -0,0 +1,50 @@
...
+----
+postgres:
+	reactive:
+	    client:
+            port: 5432
+            host: the-host
+            database: the-db
+            username: test
+            password: test
+            maxActiveConnections: 5
```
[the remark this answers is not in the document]
oshai: I think I fixed that now

oshai:
I added the fixes you request, Thanks.

Issue #9, opened 2019-05-10
[native-image] Support EntityManager generation


bufferings:
Hi, I would like to know Micronaut team's opinion to send my PR.

I have 2 questions:
1. Could I send a PR with these changes?
2. Which branch should I send PR to?

I'm thinking to send pull request as a first step for native-image generation with following changes:
https://github.com/bufferings/micronaut-sql/pull/1/files

This includes 2 changes.

1. Change to use Micronaut 1.1.0
2. Change Entity scan from Environment to BeanIntrospector

The reason why I needed to use Micronaut 1.1.0 is that BeanIntrospector doesn't exist in Micronaut 1.0.4.

graemerocher:
1. Yes please!
2. To master please

We probably want to not modify the `package-to-scan` property and maybe add an additional one that is `packages-to-use` or something. There may be cases where you still want to use package scanning.

bufferings:
Understood! At first, I will send a PR and then will modify it to use `packages-to-use` or something.

bufferings:
Close this issue as my questions are solved and created PR

Issue #8, opened 2019-04-29
Outdated micronaut version


fmayoral:
Micronaut version are misaligned for this version, see dep tree
```
[INFO] +- io.micronaut.configuration:micronaut-jdbc-hikari:jar:1.1.0:compile
[INFO] |  +- io.micronaut:micronaut-jdbc:jar:1.0.4:compile
[INFO] |  +- io.micronaut:micronaut-inject:jar:1.0.4:compile
[INFO] |  |  \- io.micronaut:micronaut-core:jar:1.0.4:compile
[INFO] |  |     \- org.reactivestreams:reactive-streams:jar:1.0.2:compile
[INFO] |  \- com.zaxxer:HikariCP:jar:3.3.1:compile
```

graemerocher:
It is not misaligned, the library still supports Micronaut 1.0.4 to align the versions correctly you should use the `micronaut-bom` for 1.1.0 which will ensure the correct transitives are used.

bufferings:
@graemerocher I understood your thought and I agree.

I'm not sure here is the right place to ask, but I have a question related 1.1.0. When I replaced `micronautVersion` to `1.1.0` in `gradle.properties` and run `./gradlew check`, a test fails:
```
Expected exception of type 'java.lang.RuntimeException', but got 'java.lang.Exception'
	at org.spockframework.lang.SpecInternals.checkExceptionThrown(SpecInternals.java:80)
	at org.spockframework.lang.SpecInternals.thrownImpl(SpecInternals.java:67)
	at io.micronaut.configuration.hibernate.jpa.JpaSetupSpec.test spring based transaction management(JpaSetupSpec.groovy:153)
Caused by: java.lang.Exception: bad things happened
	at io.micronaut.configuration.hibernate.jpa.BookService.saveError(JpaSetupSpec.groovy:226)
	at io.micronaut.context.AbstractExecutableMethod.invoke(AbstractExecutableMethod.java:144)
	at io.micronaut.aop.chain.InterceptorChain.lambda$new$1(InterceptorChain.java:89)
	at io.micronaut.aop.chain.InterceptorChain.proceed(InterceptorChain.java:146)
	at io.micronaut.spring.tx.annotation.TransactionInterceptor.intercept(TransactionInterceptor.java:81)
	at io.micronaut.aop.MethodInterceptor.intercept(MethodInterceptor.java:40)
[...]

PR #10, opened 2019-05-10
[native-image] Provide an option for `Entity` scan to use `BeanIntrospector`

bufferings:
## Background

When running as native-image, `environment.scan` doesn't work. It is one of the reasons we can't get `EntityManager` which needs `Entity` scan with the `environment.scan` in `EntityManagerFactoryBean`.

## Solution

Provide an option for `Entity` scan to use `BeanIntrospector` instead of `environment.scan`

## Changes

1. Change to use Micronaut 1.1.0 to use `BeanIntrospector`
2. Provide an option to use `BeanIntrospector` instead of `environment.scan`

## Related Issues

* https://github.com/micronaut-projects/micronaut-sql/issues/9
* https://github.com/micronaut-projects/micronaut-core/issues/1569
* https://github.com/micronaut-projects/micronaut-sql/issues/8

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-projectVersion=1.1.1.BUILD-SNAPSHOT
+projectVersion=1.2.0.BUILD-SNAPSHOT
 reactivePgClientVersion=0.11.3
 jasyncClientVersion=0.9.51
 kafkaVersion=2.1.0
@@ -11,7 +11,7 @@ commonsDbcpVersion=2.6.0
 micrometerVersion=1.1.2
 validationVersion=2.0.1.Final
 jcacheVersion=1.1.0
-micronautVersion=1.0.4
+micronautVersion=1.1.0
 micronautTestVersion=1.0.2
 protocVersion=3.5.1
 groovyVersion=2.5.6
--- a/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/EntityManagerFactoryBean.java
+++ b/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/EntityManagerFactoryBean.java
@@ -26,6 +26,8 @@ import io.micronaut.context.annotation.Factory;
 import io.micronaut.context.annotation.Parameter;
 import io.micronaut.context.annotation.Requires;
 import io.micronaut.context.env.Environment;
+import io.micronaut.core.beans.BeanIntrospection;
+import io.micronaut.core.beans.BeanIntrospector;
 import io.micronaut.core.util.ArrayUtils;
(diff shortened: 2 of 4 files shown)


bufferings:
I will modify the code not to replace `packages-to-scan` but to add new configuration such as `packages-to-use` as @graemerocher 's comment here: https://github.com/micronaut-projects/micronaut-sql/issues/9#issuecomment-491340413

bufferings:
@graemerocher , I have 2 ideas for the implementation to keep current scan featurer. Which do you think is better?

## 1. Introduce `packages-to-find`

We can use `packages-to-find` property for BeanIntrospector

```yaml
jpa:
  default:
    packages-to-find:
      - com.example.micronaut.petclinic
```

For those who prefer environment scan, they can use `packages-to-scan` same as before:

```yaml
jpa:
  default:
    packages-to-scan:
      - com.example.micronaut.petclinic
```

Restriction: They can't co-exist. We only can use one of them.

## 2. Introduce `scan-by-introspection`

```yaml
jpa:
  default:
    packages-to-scan:
      - com.example.micronaut.petclinic
    scan-by-introspection: true
```

*  `true`: use BeanIntrospector
* `false` (default value): use environment.scan

----

If anything, I'd like (2), because users don't need to understand the restriction of the two scan options. I'm glad if you kindly tell me your opinion. If you have any other options or better property names, please let me know.

graemerocher:
The latter seems good to me

graemerocher:
btw there are other PRs that can improve the experience too. The `@TypeHint` can be a PR to this project:

https://github.com/bufferings/micronaut-petclinic/blob/master/src/main/java/com/example/micronaut/petclinic/PetClinicApplication.java#L109

With all the Hibernate classes and then possibly a separate PR to `micronaut-core` for the Thymleaf classes. That way the `micronaut-hibernate-jpa` and `micronaut-views` jars will already contain the necessary type hint

Also the `native-image` metadata at https://github.com/bufferings/micronaut-petclinic/tree/master/src/main/resources/META-INF/native-image/com.example.micronaut/petclinic-application can be included with this project too.

Also we should probably change the `@Introspected` Graal support to use `allDeclaredFields` instead so it just works out of the box

bufferings:
Thank you. Yes, you're right. Actually, I'm checking Micronaut PetClinic to run with GraalVM 19.0 native image.

graemerocher:
In the end I went for a different approach and the config is:

```
jpa:
    default:
        entity-scan:
            packages:
                - 'foo.bar'
```

The `packages-to-scan` option is still possible but that enables full classpath scanning. I included documentation.

bufferings:
Looks great!

PR #18, opened 2019-07-25
update postgres-reactive to vertx-pg-client

shenzhou-6:
Postgres-reactive has been changed to vertx-pg-client
See:[https://vertx.io/docs/vertx-pg-client](https://vertx.io/docs/vertx-pg-client)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 projectVersion=1.2.0.BUILD-SNAPSHOT
-reactivePgClientVersion=0.11.3
+reactivePgClientVersion=3.8.0
 jasyncClientVersion=0.9.51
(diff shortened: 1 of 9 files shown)


BillyYccc:
Hi, I don't think you should replace the reactive-pg-client(see #11), it's still being maintained, instead you could create independent modules for the Vert.x Reactive SQL client and it also supports MySQL now.

shenzhou-6:
> Hi, I don't think you should replace the reactive-pg-client(see #11), it's still being maintained, instead you could create independent modules for the Vert.x Reactive SQL client and it also supports MySQL now.

I will resubmit.....and close this

graemerocher:
Yeah we need a module, otherwise thanks for the PR!

PR #73, opened 2019-10-19
the Reactive SQL Client dependencies clean up and upgrade from 3.8.2 to 3.8.3

### LINKED ISSUE #72: the Reactive SQL Client clean up
[the Reactive SQL Client](https://github.com/eclipse-vertx/vertx-sql-client) has been integrated into Vert.x stack, so it's unnecessary to exclude the `vertx-core` dependency in both `vertx-mysql-client` and `vertx-pg-client` modules anymore. This would ease the upgrading process of Vert.x dependencies.

BillyYccc:
Fixes #72

### DIFF
--- a/vertx-mysql-client/build.gradle
+++ b/vertx-mysql-client/build.gradle
@@ -1,13 +1,5 @@
-plugins {
-    id 'java'
-}
-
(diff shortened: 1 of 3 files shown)


graemerocher:
Thank you!

PR #107, opened 2020-02-08
Adding support for jdbi

drmaas:


jdbi/src/main/java/io/micronaut/configuration/jdbi/spring/SpringTransactionHandler.java
```diff
@@ -0,0 +1,205 @@
...
+			throw e;
+		}
+
+		didTxnRollback.remove();
+		return returnValue;
+	}
+
+	@Override
+	public <R, X extends Exception> R inTransaction(Handle handle,
+													TransactionIsolationLevel level,
```
drmaas: This appears to be github incorrectly formatting code.

graemerocher:
Thanks for the contribution! Will take a look soon

jdbi/src/main/java/io/micronaut/configuration/jdbi/JdbiFactory.java
```diff
@@ -0,0 +1,93 @@
...
+        // install all plugins with ServiceLoaders that are found on the classpath
+        jdbi.installPlugins();
+
+        // transaction handler
+        if (transactionHandler != null) {
+            jdbi.setTransactionHandler(transactionHandler);
+        }
+
+        // statement builder
+        ctx.findBean(StatementBuilderFactory.class, Qualifiers.byName(name))
```
graemerocher: This explicit call to the bean factory can be replaced by `@Parameter @Nullable StatementBuilderFactory statementBuilderFactory` as a method argument

jdbi/src/main/java/io/micronaut/configuration/jdbi/JdbiFactory.java
```diff
@@ -0,0 +1,93 @@
...
+                .ifPresent(jdbi::setStatementBuilderFactory);
+
+        // customizer - allows users to create customized configurations as defined in
+        // http://jdbi.org/apidocs/index.html?org/jdbi/v3/core/config/Configurable.html
+        ctx.findBean(JdbiCustomizer.class, Qualifiers.byName(name))
```
graemerocher: This explicit call to the bean factory can be replaced by `@Parameter @Nullable JdbiCustomizer jdbiCustomizer` as a method argument

jdbi/src/main/java/io/micronaut/configuration/jdbi/spring/SpringTransactionHandler.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.function.Consumer;
+
+/**
+ * Allows Spring Transaction to be used with Jdbi.
+ *
+ * @author Dan Maas
+ * @since 1.3.1
+ */
+@Requires(classes = PlatformTransactionManager.class)
+@EachBean(PlatformTransactionManager.class)
```
graemerocher: The Spring support is nice. I wonder would it make sense to add a dependency on `micronaut-data-tx` and include support for Micronaut's transaction management and Micronaut Data.

See https://search.maven.org/artifact/io.micronaut.data/micronaut-data-tx/1.0.0/jar

And https://github.com/micronaut-projects/micronaut-data/blob/master/data-tx/src/main/java/io/micronaut/transaction/jdbc/DataSourceTransactionManager.java

drmaas: @graemerocher would that be `instead` of using spring? It looks like micronaut's txn mgmt is mutually exclusive. If so, it might make sense to update the other libraries (like jooq) to use the same.

drmaas: I am also not sure how to test both of these in the same project due to the classpath exclusions. Would there be separate gradle modules for `jdbi-spring` and `jdbi-micronaut-data`?

drmaas: I can assume that the same would go for the jooq library

graemerocher: Yeah you make some good points. We need to update JOOQ too

I think it is probably possible to do it in a single project without splitting if requirements are added like `@Requires(property="jdbi.spring.transactions", value="true", defaultValue="true")` and then you could set `jdbi.spring.transactions` to disable spring transaction management for the scope of the tests

jdbi/src/main/java/io/micronaut/configuration/jdbi/JdbiFactory.java
```diff
@@ -0,0 +1,96 @@
...
+     * @return A {@link Jdbi}
+     */
+    @EachBean(DataSource.class)
+    public Jdbi jdbi(
+            @Parameter String name,
+            DataSource dataSource,
+            @Parameter @Nullable TransactionHandler transactionHandler,
+            @Parameter @Nullable StatementBuilderFactory statementBuilderFactory,
+            @Parameter @Nullable JdbiCustomizer jdbiCustomizer,
+            ApplicationContext ctx
```
graemerocher: You can remove this argument now since it isn't used

jdbi/src/main/java/io/micronaut/configuration/jdbi/spring/SpringTransactionHandler.java
drmaas: I'm still not sure how the Micronaut data transaction manager can be tested - it won't load if the spring transaction manager is on the classpath.

I am testing the micronaut-data transaction manager locally by toggling the gradle dependency, and for some reason rollback is not working. I can see the connection rollback happen, but the statements in the transaction get applied to the database anyways.

graemerocher: I will have a look and get back to you

graemerocher: So its cause by the lack of a transaction aware datasource proxy for the Micronaut impl. We have implemented this for the next version of Micronaut Data 1.0.1 

https://github.com/micronaut-projects/micronaut-data/pull/402

I guess to get it working for 1.0.0 of Micronaut Data you could wrap the datasource (if it is not an instance of `DelegatingDataSource` the same way as https://github.com/micronaut-projects/micronaut-data/pull/402/files#diff-ec1dcaa43f69ca2d01475a1d88349df5R24 does. Otherwise we can put off this feature until 1.0.1 is out and support it then.

drmaas: @graemerocher I think waiting until 1.0.1 is easier, and in the same feature jooq and anything else can also be upgraded to support micronaut data transactions.

hibernate-jpa/src/test/groovy/io/micronaut/configuration/hibernate/jpa/MultipleDataSourceJpaSetupSpec.groovy
```diff
@@ -73,8 +73,8 @@ class MultipleDataSourceJpaSetupSpec extends Specification{
         MutipleDataSourceJavaService javaService = applicationContext.getBean(MutipleDataSourceJavaService)
 
         expect:"Methods that retrieve the current session don't throw an exception"
-        service.testOther()
-        service.testCurrent()
+//        service.testOther()
+//        service.testCurrent()
```
graemerocher: Why have these been commented out?

jdbi/src/main/java/io/micronaut/configuration/jdbi/transaction/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Classes specific to combining JOOQ with Spring JdbcTemplate.
+ *
+ * @author Dan Maas
+ * @since 1.3.1
```
graemerocher: I have changed the version in master to `1.4.0` snapshot

Since this is a new feature it warrants a new minor release not a patch a release. Could you change this `@since` references to `1.4.0`

src/main/docs/guide/introduction.adoc
```diff
@@ -2,6 +2,10 @@ This project includes modules to support SQL Database access in Micronaut.
 
 === Release History
 
+==== 1.3.1
```
graemerocher: Change to 1.4.0

hibernate-jpa/src/test/groovy/io/micronaut/configuration/hibernate/jpa/MultipleDataSourceJpaSetupSpec.groovy
drmaas: accident

graemerocher:
Thanks for the contribution!

PR #110, opened 2020-02-11
jasync health check failing

arudge:
We started using jasync in our project and the default health check always results in a timeout.
We had to implement our own. 
I think there is an issue with the ` s.onNext(status.build());` being subscribed on. 
I wrote a failing test using the `EmbeddedServer` against the existing code.
I then modified the existing health check to pass that failing test.

I'd like to clean this all up with a :bowtie:,  but I wanted to get some feedback on using the `EmbeddedServer` for testing in this project and the overall approach.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -46,7 +46,7 @@ subprojects { Project subproject ->
     repositories {
         jcenter()
     }
-    
+
     apply plugin:"groovy"
     apply plugin:"java-library"
     apply plugin: "io.spring.nohttp"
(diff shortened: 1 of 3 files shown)


graemerocher:
All looks fine to me, any others changes you want to make?

arudge:
Great, Yes I'd like to clean things up a bit. 
I'll do that now. Thanks for the preliminary review.

graemerocher:
Will need to rebase as well. Thanks

mliberato:
I've tested on my application and it has resolved the problem here.

Will this fix arrive at 1.3.x or 1.4.x releases? Without it basically nobody can use default health endpoint unless @Replaces jasync health check bean.

@arudge it would be great to fix the typo in the class name as well. From `JaysncHealthIndicator` to `JasyncHealthIndicator`

arudge:
I'll dust this off....

arudge:
@graemerocher this is ready to go now

graemerocher:
Thanks for the contribution!

Issue #148, opened 2020-04-27
[docs] Reactive and vert.x clients are one and the same


benmccann:
https://github.com/reactiverse/reactive-pg-client redirects to https://github.com/eclipse-vertx/vertx-sql-client. The vert.x clients are reactive. There's no separate Reactive vs non-reactive ones

PR #149, opened 2020-04-27
Reactive and vert.x clients are one and the same

benmccann:
Closes https://github.com/micronaut-projects/micronaut-sql/issues/148

https://github.com/reactiverse/reactive-pg-client redirects to https://github.com/eclipse-vertx/vertx-sql-client. The vert.x clients are reactive. There's no separate Reactive vs non-reactive ones


graemerocher:
Thanks for the contribution!

PR #152, opened 2020-04-30
Upgrade SQL to Micronaut 2.x

alvarosanchez:
See https://github.com/micronaut-projects/micronaut-core/issues/3024


ilopmar:
These changes were manually applied.

sdelamo:
I am not sure if this was closed by accident. It have searched for `testImplementation "io.micronaut.cache:micronaut-cache-core"` and there are no ocurrences. I don't think this changes were applied.

PR #172, opened 2020-05-27
Improve GraalVM support for JPA (backported from 2.2.3)

ilopmar:


ilopmar:
Once this is merged we could release `1.4.1` and update the BOM in micronaut `1.3.x` branch.

PR #177, opened 2020-06-01
Enable GraalVM native image build with JOOQ

morki:


graemerocher:
Thanks for the contribution

PR #207, opened 2020-06-18
Improve jOOQ GraalVM support

ilopmar:


graemerocher:
See we may want to update micronaut launch to include this JOOQ config if the graalvm feature is selected

ilopmar:
Not really sure how we can include that. I've seen there are a few jOOQ gradle plugins (and probably same for Maven) and I used that because the user who submitted the PR used that.

PR #209, opened 2020-06-19
Add jakarta.persistence-api dependency

morki:
Needed for native image (adding `@Entity` annotation to `UpdatableRecord`s for reflection registration).

@ilopmar We should remove this from https://github.com/micronaut-graal-tests/micronaut-jooq-graal if this will be merged.

### DIFF
--- a/jooq/build.gradle
+++ b/jooq/build.gradle
@@ -19,6 +19,9 @@ dependencies {
     compileOnly "org.graalvm.nativeimage:svm"
     testCompileOnly "io.micronaut:micronaut-inject-java"
 
+    // needed for native image (adding @Entity annotation to UpdatableRecords for reflection registration)
+    implementation "jakarta.persistence:jakarta.persistence-api:2.2.2"
+
     api "io.micronaut:micronaut-jdbc:$micronautVersion"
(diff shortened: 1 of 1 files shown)


ilopmar:
@morki the changes have been included in Micronaut and the new snapshot version has been published. Even though I see the new dependency pulled in it's not working. If I remove the dependency in our test application the compilation fails with errors like
```
/home/ivan/workspaces/micronaut-graal-tests/micronaut-jooq-graal/build/generated-src/jooq/devDb/example/jooq/domain/tables/records/PetRecord.java:12: error: package javax.persistence does not exist
import javax.persistence.Column;
                        ^
```

morki:
@ilopmar I don't understand it. Did I do something wrong with declaring that dependency? Why it is not propagated? Do you have any idea, please?

ilopmar:
The problem is that it should have been `api` instead of `implementation`. At least, with `api` seems to be working now.

morki:
@ilopmar Sorry for this... :( Can you change it please?

ilopmar:
yes, don't worry.
P.S: It's not your fault, I always get it wrong :-P

PR #213, opened 2020-06-22
Scope needs to be api instead of implementation

ilopmar:


graemerocher:
Why? Does user code need to reference the JPA API when using JOOQ?

ilopmar:
Yes. When configuring jooq to enable the `jpaAnnotations`, all the generated classes have `@Entity` and related annotations. We use those to add them automatically to `reflect-configuration.json` (via GraalTypeElementVisitor) so users don't need to do anything.

PR #264, opened 2020-07-27
Dependency upgrades

micronaut-build:
Upgrades dependencies to their latest versions


sdelamo:
I am closing this since I think we should stay in 2.0.0

PR #279, opened 2020-08-10
Hibernate JPA support for additional XML mapping resources

donbeave:
This feature is equivalent to Spring Boot's feature:

```yaml
spring:
  jpa:
    mapping-resources: hibernate/custom.hbm.xml
```

With these changes, it will enable support for external mapping files with such config:

```yaml
jpa:
  default:
    mapping-resources: hibernate/custom.hbm.xml
```

Can provide a list of resources as well:

```yaml
jpa:
  default:
    mapping-resources: 
    - hibernate/custom1.hbm.xml
    - hibernate/custom2.hbm.xml
```


graemerocher:
Thanks for the contribution!

PR #286, opened 2020-08-19
UCP dependency updates

aalmiray:
UCP 19.7.0.0 has no additional dependencies.
OJDBC8 moved to test dependencies.


aalmiray:
Signed with wrong email (ugh).

PR #349, opened 2020-10-23
Compile-time Hibernate proxies

dstepanov:
The implementation uses Micronaut's AOP to generate a proxy during the compile-time.
Benefits should be:
- performance
- startup time because it looks like bytebuddy is generating some enhancements for all the classes
- Proxies can be used in GraalVM

### DIFF
--- a/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxy.java
+++ b/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxy.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.configuration.hibernate.jpa.proxy;
+
+import io.micronaut.aop.MethodInterceptor;
+import io.micronaut.core.annotation.Internal;
+import org.hibernate.proxy.HibernateProxy;
+
+/**
+ * Introduced {@link HibernateProxy} interface to register an interceptor.
+ *
+ * @author Denis Stepanov
+ * @since 3.3.0
+ */
+@Internal
+public interface IntroducedHibernateProxy extends HibernateProxy {
+
+    @SuppressWarnings("MethodName")
+    void $registerInterceptor(MethodInterceptor<Object, Object> interceptor);
+
+}
--- a/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxyAdvice.java
+++ b/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxyAdvice.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.configuration.hibernate.jpa.proxy;
+
+import io.micronaut.aop.MethodInterceptor;
+import io.micronaut.aop.MethodInvocationContext;
+import io.micronaut.context.annotation.Prototype;
+import io.micronaut.core.annotation.Internal;
+
+/**
+ * Default delegating interceptor for all {@link GenerateProxy} proxies.
+ *
+ * @author Denis Stepanov
+ * @since 3.3.0
+ */
+@Internal
+@Prototype
+public final class IntroducedHibernateProxyAdvice implements MethodInterceptor<Object, Object> {
+
+    private static final String INITIALIZE_PROXY_METHOD = "$registerInterceptor";
+
+    private MethodInterceptor<Object, Object> interceptor;
+
+    @Override
+    public Object intercept(MethodInvocationContext<Object, Object> context) {
+        if (interceptor == null) {
+            if (INITIALIZE_PROXY_METHOD.equals(context.getMethodName())) {
+                interceptor = (MethodInterceptor<Object, Object>) context.getParameterValues()[0];
+                return null;
+            }
+            return context.proceed();
+        }
+        return interceptor.intercept(context);
+    }
+
+}
--- a/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxyFactory.java
+++ b/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxyFactory.java
@@ -82,7 +82,7 @@ final class IntroducedHibernateProxyFactory implements ProxyFactory {
     public HibernateProxy getProxy(Serializable id, SharedSessionContractImplementor session) throws HibernateException {
         if (beanDefinition == null) {
             beanDefinition = beanContext.findProxyTargetBeanDefinition(persistentClass, null)
-                    .orElseThrow(() -> new HibernateException("Cannot find a proxy class, please annotate " + persistentClass + " with @IntroduceHibernateProxy."));
+                    .orElseThrow(() -> new HibernateException("Cannot find a proxy class, please annotate " + persistentClass + " with @GenerateProxy."));
         }
         LazyInitializer lazyInitializer = new IntroducedHibernateProxyLazyInitializer(entityName, persistentClass, id, session);
         Object proxyTargetBean = beanContext.getBean(beanDefinition);
--- a/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxyFactory.java
+++ b/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxyFactory.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.configuration.hibernate.jpa.proxy;
+
+import io.micronaut.context.BeanContext;
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.inject.BeanDefinition;
+import io.micronaut.inject.ExecutableMethod;
+import org.hibernate.HibernateException;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.proxy.HibernateProxy;
+import org.hibernate.proxy.LazyInitializer;
+import org.hibernate.proxy.ProxyFactory;
+import org.hibernate.type.CompositeType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.Serializable;
+import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * Implementation of Hibernate's {@link ProxyFactory}.
+ * Compile time proxy will be used if one is added by annotating an entity with {@link GenerateProxy}.
+ *
+ * @author Denis Stepanov
+ * @since 3.3.0
+ */
+@Internal
+final class IntroducedHibernateProxyFactory implements ProxyFactory {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(IntroducedHibernateProxyFactory.class);
+    private static final Set<Class<?>> EXPECTED_INTERFACES = Collections.singleton(HibernateProxy.class);
+    private static final String GET_HIBERNATE_LAZY_INITIALIZER = "getHibernateLazyInitializer";
+
+    private final BeanContext beanContext;
+
+    private String entityName;
+    private Class<?> persistentClass;
+    private CompositeType componentIdType;
+    private Method getIdentifierMethod;
+    private Method setIdentifierMethod;
+
+    private BeanDefinition<?> beanDefinition;
+
+    public IntroducedHibernateProxyFactory(BeanContext beanContext) {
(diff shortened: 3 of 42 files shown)


dstepanov:
@graemerocher What do you think?

ilopmar:
Independently of Graeme's thoughts, we need to test that this still works with GraalVM before merging it. Once Graeme reviews it and approves the PR I will test it, and if everything works I'll merge it.

dstepanov:
@ilopmar What is the way to test it on GraalVM?

dstepanov:
@ilopmar If I'm correct proxies don't work on GraalVM at the moment so this cannot break it.

ilopmar:
Build the different applications in https://github.com/micronaut-graal-tests/micronaut-data-jpa-graal and https://github.com/micronaut-graal-tests/micronaut-data-jdbc-graal (for all the databases) with the local snapshot of micronaut-sql, and a GraalVM version compiled from master branch.

ilopmar:
Actually we don't need to test the Data JDBC applications because you've only modified the JPA stuff. But I think we need to test it because you've modified `HibernateSubstitutions`

graemerocher:
Generally with the new updates to the Gradle plugin for Micronaut that includes the ability to test native images I think where possible we should move native image tests to the subprojects themselves

dstepanov:
Would be nice if the plugin supported test containers. I can imagine test containers can be usered by the plugin to start the container and propagate the address+ip using env.

graemerocher:
I think that is already doable now, you would have to implement `TestPropertyProvider` and spin up the container and pass the configuration then micronaut test would create the native image process and the settings from the test will be passed to it and the native image can connect to the container

dstepanov:
I have added a GraalVM test project, originaly I wanted to include more DB drivers and create mutliple DB specific tests but it looks like I cannot make MySQL or H2 drivers work, those I tried besides Postgres.
I have tried to add:
```
nativeImage {
    args('--report-unsupported-elements-at-runtime')
}
```
Native image reports errors on Java 8 and Java 11, I suspect it might have something to do with `hibernate-jpa` being referenced as a project.

@graemerocher Is intentional that env from the test aren't propagated to the native test app? It makes things more complicated if you want to support ordinary + native test, it would be nice to enable the same "remote" test for ordinary JVM test runner.

dstepanov:
Actually the problem is only when postgres and mysql drivers are together on the classpath, `nativeimage` fails with:

`
Error: Incompatible change of initialization policy for java.sql.DriverManager: trying to change BUILD_TIME from feature io.micronaut.core.graal.AutomaticFeatureUtils.lambda$initializeAtBuildTime$0 to RUN_TIME from feature io.micronaut.core.graal.AutomaticFeatureUtils.lambda$initializeAtRunTime$1
`

graemerocher:
@dstepanov I believe the test properties are propagated. See https://github.com/micronaut-projects/micronaut-test/blob/master/test-core/src/main/java/io/micronaut/test/support/server/TestExecutableEmbeddedServer.java#L173

dstepanov:
@graemerocher Yes, but not environments see "-Dmicronaut.environments=test"`. I was trying to have DB initializer that adds testing data, the problem was that there are two contexts in the case of the native test, and DB was getting filled twice.  No way to exclude the runner context because it has all the properties and envs.

graemerocher:
perhaps we need a `native` environment that you can exclude or perhaps we shouldn't set the environment in the forked server 🤔

dstepanov:
Looks like `JdbcFeature` was adding `java.sql.DriverManager` with a different initialize stage, now postgres and mysql tests run together.

@graemerocher Probably, I would prefer to have `@MicronautTest(remote=true)` that would run separate context for ordinary tests and it will only pick remotes for native tests. That would be much better to debug and understand separate contexts tests.

graemerocher:
yeah the issue it that would need to be part of the plugin since they only way you could get `remote=true` to work the same is to isolate the test classpath and runtime classpath since what you would need to do is run the application without any test modifications / classpath changes

dstepanov:
I have added more DBs, almost all of them from https://github.com/micronaut-graal-tests/micronaut-data-jpa-graal except h2 and MSSQL which have drivers that don't compile with GraalVM 20.2.

@graemerocher @ilopmar WDYT? Are you planning to modify the pipline to execute `testNativeImage` or is it going to be just for manual validation?

graemerocher:
We should probably do what I did on the r2dbc subproject. See https://github.com/micronaut-projects/micronaut-r2dbc/blob/master/.github/workflows/graalvm.yml

hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroduceHibernateProxy.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @author Denis Stepanov
+ */
+@Around(proxyTarget = true)
+@Introduction(interfaces = IntroducedHibernateProxy.class)
+@Type(IntroducedHibernateProxyAdvice.class)
+@Documented
+@Retention(RUNTIME)
+@Target({ElementType.TYPE})
+public @interface IntroduceHibernateProxy {
```
graemerocher, line 40: Maybe rename to `@GenerateProxy`

hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroduceHibernateProxy.java
```diff
@@ -0,0 +1,41 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Annotated Entity class will have a compile time Hibernate proxy.
+ *
+ * @author Denis Stepanov
```
graemerocher, line 32: needs `@since`

hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxy.java
```diff
@@ -0,0 +1,31 @@
...
+
+import io.micronaut.aop.MethodInterceptor;
+import org.hibernate.proxy.HibernateProxy;
+
+/**
+ * Introduced {@link HibernateProxy} interface to register an interceptor.
+ *
+ * @author Denis Stepanov
+ */
+public interface IntroducedHibernateProxy extends HibernateProxy {
```
graemerocher, line 26: Needs `@since` and annotate with `@Internal`

hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxyAdvice.java
```diff
@@ -0,0 +1,46 @@
...
+import io.micronaut.aop.MethodInvocationContext;
+import io.micronaut.context.annotation.Prototype;
+
+/**
+ * Default delegating interceptor for all {@link IntroduceHibernateProxy} proxies.
+ *
+ * @author Denis Stepanov
+ */
+@Prototype
+public class IntroducedHibernateProxyAdvice implements MethodInterceptor<Object, Object> {
```
graemerocher, line 28: Needs `@since` and annotate with `@Internal`

hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroduceHibernateProxy.java
graemerocher: Would also probably annotate with `@Experimental` for now

hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxyFactory.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Set;
+
+/**
+ * Implementation of Hibernate's {@link ProxyFactory}.
+ * Compile time proxy will be used if one is added by annotating an entity with {@link IntroduceHibernateProxy}.
+ *
+ * @author Denis Stepanov
+ */
+@Internal
+class IntroducedHibernateProxyFactory implements ProxyFactory {
```
graemerocher, line 43: Declare `final` and add `@since`

hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxyFactoryFactory.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.Arrays;
+
+/**
+ * Implementation of Hibernate's {@link ProxyFactoryFactory} that creates {@link IntroducedHibernateProxyFactory} with {@link BeanContext}.
+ *
+ * @author Denis Stepanov
+ */
+@Internal
+class IntroducedHibernateProxyFactoryFactory implements ProxyFactoryFactory {
```
graemerocher, line 35: Declare `final` and add `@since`

hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxyFactoryFactory.java
```diff
@@ -0,0 +1,52 @@
...
+                .getService(ManagedBeanRegistry.class)
+                .getBean(BeanContext.class)
+                .getBeanInstance();
+        return new IntroducedHibernateProxyFactory(beanContext);
+    }
+
+    @Override
+    public BasicProxyFactory buildBasicProxyFactory(Class superClass, Class[] interfaces) {
+        // Fallback?
+        throw new HibernateException("IntroducedHibernateProxyFactoryFactory doesn't support to generate a BasicProxy for type " + superClass + " and interfaces " + Arrays.toString(interfaces) + ". Enable a different BytecodeProvider.");
```
graemerocher, line 49: Error message needs to be better here, maybe something like:

```
"Proxying of basic type " + superClass + " and interfaces " + Arrays.toString(interfaces) + " not supported. Micronaut only supports generating proxies for POJOs and non-basic types. Disable Micronaut compile-time proxies or remove basic type proxy to proceed."
```

hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntroducedHibernateProxyLazyInitializer.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.io.Serializable;
+
+/**
+ * Basic {@link org.hibernate.proxy.LazyInitializer}.
+ *
+ * @author Denis Stepanov
+ */
+@Internal
+class IntroducedHibernateProxyLazyInitializer extends AbstractLazyInitializer {
```
graemerocher, line 30: Declare `final` add `@since`

hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/proxy/IntrospectedHibernateBytecodeProvider.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Compile-time proxies implementation of Hibernate's {@link BytecodeProvider}.
+ * Implementation provides custom implementation of {@link ProxyFactoryFactory} and disables bytecode enhancer.
+ *
+ * @author Denis Stepanov
+ */
+@Singleton
+@Internal
+public class IntrospectedHibernateBytecodeProvider implements BytecodeProvider {
```
graemerocher, line 41: Add `@since`

graemerocher (changes requested):
Few other changes, also will need documentation for this feature.

dstepanov:
@graemerocher Should it be since 3.3.0?

graemerocher:
Yes

> @graemerocher Should it be since 3.3.0?

Yes

test-graalvm-hibernate-jpa/build.gradle
```diff
@@ -0,0 +1,114 @@
...
+
+        testImplementation "io.micronaut.test:micronaut-test-core:$micronautTestVersion"
+        testImplementation platform("org.testcontainers:testcontainers-bom:$testContainersVersion")
+        testImplementation "org.testcontainers:spock"
+        testImplementation "org.testcontainers:jdbc"
+    }
+
+    nativeImage {
+        args("-H:+TraceClassInitialization")
+        args("-H:IncludeResources=logback.xml|application.yml")
```
ilopmar, line 62: This is shouldn't be necessary. The Graal type element visitor includes automatically all resources.

dstepanov: Somehow it doesn't work, you can try to remove it and run `./gradlew :test-graalvm-hibernate-jpa:h2:testNativeImage`

dstepanov:
I had to abandon my original idea to bundle all the drivers inside one image, that doesn't work because build time initialization of `java.sql.DriverManager` makes all other drivers to initialize at the build time and MySql somehow doesn't support it.

Now there is a Gradle multi module project to test each db separately, each subproject extends shared sources which makes IntelliJ a bit confused bit otherwise it works.

I also managed to include all the DB tests: h2, mysql, mssql, postgres, mariadb and oracle. Looks like all GraalVM tests are passing for Java 8 but for Java 11 h2 native image fails.

test-graalvm-hibernate-jpa/build.gradle
ilopmar: I see it now.

This is the output of an application that works:

```
> Task :compileJava
Note: Writing native-image.properties file to destination: META-INF/native-image/example.micronaut/basic-app/native-image.properties
Note: Writing reflection-config.json file to destination: META-INF/native-image/example.micronaut/basic-app/reflection-config.json
Note: Writing resource-config.json file to destination: META-INF/native-image/example.micronaut/basic-app/resource-config.json
```

Note that the `resource-config.json` is created.


But for this project, the file is not created:

```
> Task :test-graalvm-hibernate-jpa:h2:compileJava
Note: Writing native-image.properties file to destination: META-INF/native-image/io.micronaut.example/h2/native-image.properties
Note: Writing reflection-config.json file to destination: META-INF/native-image/io.micronaut.example/h2/reflection-config.json
Note: Creating bean classes for 6 type elements
```

I can't debug the type element visitor to see why, it never stops on my breaking point. I suspect that maybe it's because of the multiproject build can't find the `resources` directory: https://github.com/micronaut-projects/micronaut-core/blob/69ec24c34aebc244209f4e2ced47af694a4d05a1/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java#L323

ilopmar:
> Now there is a Gradle multi module project to test each db separately, each subproject extends shared sources which makes IntelliJ a bit confused bit otherwise it works.

Is there any other way to configure this? It's annoying to see this in IntelliJ and also not being able to run those tests from there, only the H2. I know that those tests are only for GraalVM, but still...

image



> I also managed to include all the DB tests: h2, mysql, mssql, postgres, mariadb and oracle. Looks like all GraalVM tests are passing for Java 8 but for Java 11 h2 native image fails.

I still don't know why this happened. When GraalVM 20.2 was the master branch on graalvm repository, all our tests passed, but somehow, once it was released some of them doesn't pass.

---

A couple of additional question @dstepanov:
- Do you plan to add another similar sub-project for the Micronaut Data JDBC applications?
- And what about JPA but with Micronaut Data?

test-graalvm-hibernate-jpa/build.gradle
dstepanov: Looks like the code check for the resources dir in the current directory, but in this case, it's in a different location.

ilopmar: I see :+1:

dstepanov:
@ilopmar I was trying to make a subproject but somehow Hibernate didn't work, not sure what was that about, error was saying that it cannot find the id in metadata, not sure what is happaning with Hiberante in the context of GraalVM.

I think because Hibernate is using JDBC internally it's enought to have just those tests. The data project would need to have separate tests.

Any idea how to disable h2 tests just on GraalVM Java 11?

dstepanov:
@ilopmar I have realized what was missing to make it a subproject

dstepanov:
All green 💚

dstepanov:
Added documentation

graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #350, opened 2020-10-25
Add ability to configure jOOQ ConverterProvider 

morki:
and create default implementation using builtin Jackson for JSON and JSONB types (must me configured)

https://www.jooq.org/doc/3.14/manual/sql-execution/fetching/converter-provider/

### DIFF
--- a/jooq/build.gradle
+++ b/jooq/build.gradle
@@ -28,6 +28,7 @@ dependencies {
     compileOnly "org.springframework:spring-jdbc:$springVersion"
     compileOnly "io.micronaut.data:micronaut-data-tx:$micronautDataVersion"
     compileOnly "org.simpleflatmapper:sfm-reflect:8.2.3"
+    compileOnly "com.fasterxml.jackson.core:jackson-databind"
 
     testRuntimeOnly project(":jdbc-tomcat")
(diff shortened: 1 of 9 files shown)


morki:
It is related to new feature in jOOQ 3.14 - transforming database JSON values, by default jOOQ use reflection and creates default ObjectMapper if Jackson is on classpath. But of course this default ObjectMapper is not configured by Micronaut, so is not using introspection and configured modules (like KotlinModule in our case).

So this PR adds:
- missing ability to replace ConverterProvider (needed anyway)
- default implementation configurable by application configuration to create Converter with built-in ObjectMapper configured by Micronaut

graemerocher:
Does this help with the native image problems?

morki:
I some way, it does not resolve https://github.com/micronaut-graal-tests/micronaut-jooq-graal/issues/9 but It uses `Introspection` for serialization and deserialization, so no additional reflection configuration is needed.

morki:
@graemerocher is this something that can be accepted or do I have to split those functionalities to:

1. add option to configure ConverterProvider (really needed)
2. default implementation using built-in ObjectMapper (can be done on user side)

graemerocher:
Thanks for the contribution!

PR #365, opened 2020-11-12
Use correct driver for MySQL

### LINKED ISSUE #364: com.mysql.jdbc.Driver is deprecated
Autodetection of the database driver is defaulting to the deprecated version of the MySQL driver:

https://github.com/micronaut-projects/micronaut-sql/blob/master/jdbc/src/main/java/io/micronaut/jdbc/JdbcDatabaseManager.java#L42

This should be replaced with `com.mysql.cj.jdbc.Driver`

graemerocher:
Fixes #364


graemerocher:
The driver also needs 

```
nativeImage {
    args('-H:+AddAllCharsets')
}
```

But that can't be added to the automatic feature as far as I know

graemerocher:
Maybe could be added to the Micronaut launch template

ilopmar:
I was already using the new driver for the GraalVM test applications: https://github.com/micronaut-graal-tests/micronaut-data-jdbc-graal/blob/f04a7cb135a65f6a8eb0b53d69b07e739c2708da/src/main/resources/application.yml#L10

In any case I'll try this PR locally with GraalVM to make sure everything works.

ilopmar (approved):
Tested with 20.3.0-dev and worked properly.

ilopmar:
> The driver also needs
> 
> ```
> nativeImage {
>     args('-H:+AddAllCharsets')
> }
> ```

It works for me when using MySQL without this. It is necessary to add it when using MSSQL Server: https://github.com/micronaut-graal-tests/micronaut-data-jdbc-graal/blob/2.2.x_sqlserver/src/main/resources/META-INF/native-image/example.jdbc/example-jdbc/native-image.properties#L4

> But that can't be added to the automatic feature as far as I know

Yes, it's not possible. I asked about it sometime ago on their Slack and they confirmed that it wasn't possible :-(

PR #378, opened 2020-11-18
Switch GraalVM testing to 20.3

dstepanov:


.github/workflows/graalvm.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        graalvm: ['20.2.0.java8', '20.2.0.java11']
+        graalvm: ['20.3.0.java8', '20.3.0.java11']
```
ilopmar, line 17: This should be done in the [project template](https://github.com/micronaut-projects/micronaut-project-template/blob/master/.github/workflows/graalvm.yml#L17) because then it would be propagated automatically to all the modules

dstepanov: Yeah probably, this also removes the predicate to skip h2 tests on Java 11

ilopmar: I mean, you can leave it here if you need it for this PR, it's not a problem at all :-)

Just keep in mind that modifications of the workflows in the projects would be overridden in the future unless you also change the template project.

dstepanov: Yeah that's fine, any idea why builds fail on the publish test reports stage?

graemerocher:
Already applied via dependency updates PR

dstepanov:
Great! Do we still need to disable h2 tests on GraalVM Java11?

PR #394, opened 2020-12-02
Allow OJDBC secure connection

graemerocher:


graemerocher:
@ilopmar could you review this and then we need to do another release, if you want I can do the release?

ilopmar:
I don't have an Oracle instance with security enabled so can't test it :(

graemerocher:
Ok well I tested it and works

Issue #433, opened 2021-03-11
If no default datasource entities are defined, no datasource entities will be processed


rawilder:
Thanks for reporting an issue, please review the task list below before submitting the
issue. Your issue report will be closed if the issue is incomplete and the below tasks not completed.

NOTE: If you are unsure about something and the issue is more of a question a better place to ask questions is on Stack Overflow (https://stackoverflow.com/tags/micronaut) or Gitter (https://gitter.im/micronautfw/). DO NOT use the issue tracker to ask questions.

### Task List

- [x] Steps to reproduce provided
- [x] Stacktrace (if present) provided
- [x] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. Declare a default and custom datasource
2. Declare the JPA properties for custom datasource
3. Create an entity
3. Create a repository that uses the custom datasource and entity
4. Create a controller that uses the repository
5. Create a get request that gets all entities from the repository
6. Run the application
7. Call the endpoint

### Expected Behaviour

It should return whatever entities you have in your datasource (empty list in the example below)

### Actual Behaviour

Throws the following exception:
```
Unexpected error occurred: No backing RepositoryOperations configured for repository. Check your configuration and try again
io.micronaut.context.exceptions.ConfigurationException: No backing RepositoryOperations configured for repository. Check your configuration and try again
        at io.micronaut.data.intercept.DataIntroductionAdvice.findInterceptor(DataIntroductionAdvice.java:124)
        at io.micronaut.data.intercept.DataIntroductionAdvice.intercept(DataIntroductionAdvice.java:78)
        at io.micronaut.aop.chain.MethodInterceptorChain.proceed(MethodInterceptorChain.java:82)
        at com.example.TestEntityRepository$Intercepted.findAll(Unknown Source)
        at com.example.TestEntityController.getAllTestEntities(TestEntityController.kt:12)
        at com.example.$TestEntityControllerDefinition$$exec1.invokeInternal(Unknown Source)
        at io.micronaut.context.AbstractExecutableMethod.invoke(AbstractExecutableMethod.java:146)
        at io.micronaut.context.DefaultBeanContext$4.invoke(DefaultBeanContext.java:474)
        at io.micronaut.web.router.AbstractRouteMatch.execute(AbstractRouteMatch.java:255)
        at io.micronaut.web.router.RouteMatch.execute(RouteMatch.java:118)
[...]

PR #434, opened 2021-03-13
Improve JPA configuration

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-sql/issues/433

This should target major release as it contains not-backward-compatible changes and I don't think there is a way around it.
The test is supposed to test all possible configuration cases.


graemerocher:
@dstepanov can you outline the nature of the changes. If there are changes they need to be documented somewhere

graemerocher:
Incorporated into https://github.com/micronaut-projects/micronaut-sql/pull/469

Issue #497, opened 2021-10-12
Exception in log stream when closing jdbc-ucp pool


pgressa:
### Expected Behavior

When the application shuts down the UCP connection pool is closed.

### Actual Behaviour

The log stream contains;
```
15:55:27.097 [Test worker] DEBUG io.micronaut.context.DefaultBeanContext - Found no possible candidate beans of type [beanPreDestroyEventListener] for qualifier: null 
15:55:27.097 [Test worker] DEBUG io.micronaut.configuration.jdbc.ucp.DatasourceFactory - Closing connection pool named: null
15:55:27.117 [Test worker] WARN io.micronaut.configuration.jdbc.ucp.DatasourceFactory - Error closing data source [oracle.ucp.jdbc.PoolDataSourceImpl@194853d2]: The Universal Connection Pool cannot be null
oracle.ucp.UniversalConnectionPoolException: The Universal Connection Pool cannot be null
	at oracle.ucp.util.UCPErrorHandler.newUniversalConnectionPoolException(UCPErrorHandler.java:369)
	at oracle.ucp.util.UCPErrorHandler.newUniversalConnectionPoolException(UCPErrorHandler.java:337)
	at oracle.ucp.util.UCPErrorHandler.newUniversalConnectionPoolException(UCPErrorHandler.java:351)
	at oracle.ucp.admin.UniversalConnectionPoolManagerBase.destroyConnectionPool(UniversalConnectionPoolManagerBase.java:509)
	at io.micronaut.configuration.jdbc.ucp.DatasourceFactory.close(DatasourceFactory.java:75)
	at io.micronaut.configuration.jdbc.ucp.$DatasourceFactory$Definition.dispose(Unknown Source)
	at io.micronaut.inject.DisposableBeanDefinition.dispose(DisposableBeanDefinition.java:44)
	at io.micronaut.context.DefaultBeanContext$BeanDisposingRegistration.close(DefaultBeanContext.java:4222)
	at io.micronaut.context.DefaultBeanContext.disposeBean(DefaultBeanContext.java:1173)
	at io.micronaut.context.DefaultBeanContext.stop(DefaultBeanContext.java:298)
	at io.micronaut.context.DefaultApplicationContext.stop(DefaultApplicationContext.java:186)
	at io.micronaut.context.DefaultApplicationContext.stop(DefaultApplicationContext.java:50)
	at io.micronaut.context.LifeCycle.close(LifeCycle.java:61)
[...]

PR #498, opened 2021-10-13
fix: ucp pool startup and configuration from calculated settings

pgressa:
Fixes: #497


jdbc-ucp/src/main/java/io/micronaut/configuration/jdbc/ucp/DatasourceConfiguration.java
```diff
@@ -183,4 +185,44 @@ public String getConfiguredValidationQuery() {
...
 
+
+    /**
+     * Configures the missing properties of the data source from the calculated settings.
+     *
+     * @since 4.0.2
+     */
+    @PostConstruct
+    public void initialize() {
+        if (Objects.equals(getConfiguredDriverClassName(), "")) {
```
burtbeckwith, line 196:
```suggestion
        if ("".equals(getConfiguredDriverClassName()) {
```

PR #493, opened 2021-10-04
[sql] Update common files for branch master

micronaut-build:
Update common files

PR #499, opened 2021-10-13
updates from template project (enable JDK 16 CI)

burtbeckwith:


pgressa:
@burtbeckwith this duplicates https://github.com/micronaut-projects/micronaut-sql/pull/493

PR #512, opened 2021-11-17
use collect for thread safe access

rawilder:


rawilder:
I was noodling on this a bit. The synchronizedSet is the least code change but there's also a solution in which we leave entities untouched and instead use `addAll` with `.collect(Collectors.toSet())` instead of `.forEach(entities::add)`. Any thoughts on that?

graemerocher:
`.collect(Collectors.toSet())` would be better

rawilder:
@graemerocher looks like the gradle download of one of the resources is experiencing a 403. I was seeing this locally as well.

rawilder:
Any update with regards to the build error?

```
Could not GET 'dl.bintray.com/micronaut/core-releases-local/io/micronaut/build/micronaut-gradle-plugins/2.0.11/micronaut-gradle-plugins-2.0.11.pom'. Received status code 403 from server: Forbidden
```

rawilder:
@graemerocher any updates?

graemerocher:
looks like build plugins need updating: https://github.com/micronaut-projects/micronaut-sql/blob/master/build.gradle#L7

@melix could you assist in getting this build up-to-date?

melix:
I will take a look.

graemerocher:
Replaced by https://github.com/micronaut-projects/micronaut-sql/pull/524

PR #524, opened 2021-12-17
use collect for thread safe access - rebased

### LINKED ISSUE #511: JpaConfiguration::findEntities is using a non-thread safe HashSet
### Expected Behavior

Multiple packages are scanned and all entities are recognized.

### Actual Behaviour

Not all entities are recognized when multiple packages are specified in entity-scan. 

I'm having a hard time validating without actually cloning and referencing a local copy version of the library, but from my understanding of HashSet, it is a non-thread safe data structure and the parallel stream being used might be an unsafe pairing. 

https://github.com/micronaut-projects/micronaut-sql/blob/3.1.x/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/JpaConfiguration.java#L284

https://github.com/micronaut-projects/micronaut-core/blob/3.1.x/core/src/main/java/io/micronaut/core/io/scan/AnnotationScanner.java#L136

From the [javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Stream.html#forEach(java.util.function.Consumer)) of `Stream::forEach` (which is used [here](https://github.com/micronaut-projects/micronaut-sql/blob/3.1.x/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/JpaConfiguration.java#L288))

> For any given element, the action may be performed at whatever time and in whatever thread the library chooses. If the action accesses shared state, it is responsible for providing the required synchronization.

### Steps To Reproduce

1. Create a micronaut app with entities in different packages
2. Specify the packages separately in a list
3. Write a basic kotest MicronautTest
4. Drop a debug in findEntities and watch how many entities are returned

### Environment Information

Jdk 11
MacOS

### Example Application

_No response_

### Version

3.1.4

graemerocher:
Fixes #511

### DIFF
--- a/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/JpaConfiguration.java
+++ b/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/JpaConfiguration.java
@@ -358,7 +358,7 @@ public class JpaConfiguration {
          * @return The entities
          */
         public Collection<Class<?>> findEntities() {
-            Collection<Class<?>> entities = new HashSet<>();
+            Collection<Class<?>> entities = Collections.synchronizedSet(new HashSet<>());
             if (isEnabled()) {
                 if (ArrayUtils.isNotEmpty(packages)) {
                     environment.scan(Entity.class, packages).forEach(entities::add);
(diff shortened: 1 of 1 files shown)


graemerocher:
@rawilder this PR is rebased to target master which we will release soon

rawilder:
Awesome! Thank you

graemerocher:
Thanks for the contribution!

PR #535, opened 2021-12-22
Use jdbi3-bom platform instead of jdbi-core

MT-Jacobs:
Correct me if I'm wrong on how bom files work - this is a new use case for me - but we can use a platform here to establish the jdbi3 dependencies rather than manually establishing all of the jdbi versions ourselves.

This way I can avoid manually defining the version number for any of the jdbi modules I want to use.

### DIFF
--- a/sql-bom/build.gradle
+++ b/sql-bom/build.gradle
@@ -9,8 +9,6 @@ micronautBom {
 }
 
 dependencies {
-    // jdbi
-    implementation platform("org.jdbi:jdbi3-bom:$jdbiVersion")
     
     constraints {
 
@@ -45,8 +43,12 @@ dependencies {
         api "org.mariadb.jdbc:mariadb-java-client:$mariadbDriverVersion"
         api "mysql:mysql-connector-java:$mysqlDriverVersion"
 
+
+        // jdbi
+        platform("org.jdbi:jdbi3-bom:$jdbiVersion")
+
         // Will remove these - just want to chat about it first
-        //api "org.jdbi:jdbi3-core:$jdbiVersion"
+        api "org.jdbi:jdbi3-core"
         //api "org.jdbi:jdbi3-sqlobject:$jdbiVersion"
         //api "org.jdbi:jdbi3-guava:$jdbiVersion"
         //api "org.jdbi:jdbi3-jodatime2:$jdbiVersion"
--- a/sql-bom/build.gradle
+++ b/sql-bom/build.gradle
@@ -9,6 +9,9 @@ micronautBom {
 }
 
 dependencies {
+    // jdbi
+    implementation platform("org.jdbi:jdbi3-bom:$jdbiVersion")
+    
     constraints {
 
         // vertx client
@@ -42,9 +45,6 @@ dependencies {
         api "org.mariadb.jdbc:mariadb-java-client:$mariadbDriverVersion"
         api "mysql:mysql-connector-java:$mysqlDriverVersion"
 
-        // jdbi
-        api platform("org.jdbi:jdbi3-bom:$jdbiVersion")
-        
         // Will remove these - just want to chat about it first
         //api "org.jdbi:jdbi3-core:$jdbiVersion"
         //api "org.jdbi:jdbi3-sqlobject:$jdbiVersion"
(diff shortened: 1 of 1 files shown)


sql-bom/build.gradle
```diff
@@ -42,7 +43,22 @@ dependencies {
         api "org.mariadb.jdbc:mariadb-java-client:$mariadbDriverVersion"
         api "mysql:mysql-connector-java:$mysqlDriverVersion"
 
+
         // jdbi
-        api "org.jdbi:jdbi3-core:$jdbiVersion"
+        platform("org.jdbi:jdbi3-bom:$jdbiVersion")
+
+        // Will remove these - just want to chat about it first
+        api "org.jdbi:jdbi3-core"
```
MT-Jacobs, line 51: Doesn't blow up, therefore the platform is being respected.

sql-bom/build.gradle
```diff
@@ -9,6 +9,8 @@ micronautBom {
 }
 
 dependencies {
+    api platform("org.jdbi:jdbi3-bom:$jdbiVersion")
```
MT-Jacobs, line 12: Proper way of consuming a platform per https://docs.gradle.org/current/userguide/java_platform_plugin.html#sec:java_platform_consumption

MT-Jacobs:
@melix, could I have you take a look at this?

melix:
I'll take a look, but I need to double check what would be the behavior with Maven, as I _think_ that doing so would force Maven users to import the `jdbi3-bom` themselves.

MT-Jacobs:
It seems that `micronaut-core` does something similar with `testcontainers-bom` -

https://github.com/micronaut-projects/micronaut-core/blob/9c60855884c6241fd967a5f8d1da364610020739/gradle/libs.versions.toml#L186

Ah! You last touched that - that would explain why this was assigned to you.

https://github.com/micronaut-projects/micronaut-core/pull/6199

melix:
I understand why my review was required. I last touched the file because I refactored the build :) I didn't forget about you.

sql-bom/build.gradle
melix: :+1:

sql-bom/build.gradle
melix: The comments below should be removed. There will apparently be a difference: currently the `micronaut-sql` BOM would produce a constraint for `org.jdbi:jdbi3-oracle12:$jdbiVersion`. For the current version, which is `3.0.21`, as per https://github.com/micronaut-projects/micronaut-sql/blob/df34ee6e8e0b049517e38f103c13e3005a9e50e6/gradle.properties#L23 and confirmed by https://ge.micronaut.io/s/xfmwthgikr56s/dependencies?dependencies=jdbi&expandAll&focusedDependency=WzgsMSwxNDg0LFswLDAsWzFdXV0.

However the 3.0.21 BOM doesn't contain an entry for `oracle12`: https://search.maven.org/artifact/org.jdbi/jdbi3-bom/3.20.1/pom

So we need to keep the `api` constraint on `api "org.jdbi:jdbi3-oracle12:$jdbiVersion"` (the other modules are all present).

However, it seems that `jdbi3-oracle12` is _already_ broken since there's no such thing as `3.0.21`: https://repo1.maven.org/maven2/org/jdbi/jdbi3-oracle12/ (it seems to be versioned independently of the rest of the project). So either we need to use a specific version of this dependency, or we simply remove it from the BOM. @graemerocher wdyt?

melix (changes requested):
Sorry for the late review. I think the changes are fine, but there's one little inconsistency that we need to figure out.

sql-bom/build.gradle
MT-Jacobs: Was `jdbi3-oracle12` ever included here? Everything below is a comment that I added as a reference, not an old part of the BOM.

MT-Jacobs: Removed the comments to make this more clear.

MT-Jacobs:
Hi, are there any remaining action items I need to complete for this PR to be approved?

melix (approved):

graemerocher:
Thanks for the contribution!

PR #547, opened 2022-01-20
fix: use method to access delegate

pgressa:


pgressa:
@graemerocher or @burtbeckwith  .. this is needed to fix https://github.com/micronaut-projects/micronaut-oracle-cloud/issues/325 , so the `@PostConstruct` in `io.micronaut.configuration.jdbc.ucp.DatasourceConfiguration` is possible to intercept by `BeanInitializedEventListener`

PR #571, opened 2022-03-11
Next minor

timyates:
- Update version to 4.2.0-SNAPSHOT
- Update to Micronaut 3.3.4
- Update branch to 3.4.x
- Update Groovy version to 3.0.10
- Clean up unrequired dependencies

### DIFF
--- a/buildSrc/src/main/groovy/io.micronaut.build.internal.sql-module.gradle
+++ b/buildSrc/src/main/groovy/io.micronaut.build.internal.sql-module.gradle
@@ -2,10 +2,6 @@ plugins {
     id 'io.micronaut.build.internal.module'
 }
 
-dependencies {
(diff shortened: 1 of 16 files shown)


buildSrc/src/main/groovy/io.micronaut.build.internal.sql-module.gradle
```diff
@@ -2,10 +2,6 @@ plugins {
     id 'io.micronaut.build.internal.module'
 }
 
-dependencies {
-    implementation 'jakarta.inject:jakarta.inject-api:2.0.1'
-}
-
```
timyates, line 8: I don't *think* we need this anymore 🤔

sdelamo (approved):

PR #574, opened 2022-03-16
Fix BOM generation

melix:
This fixes the root cause for https://github.com/micronaut-projects/micronaut-core/issues/7068
which is that the BOM we generated contained a constraint without version.

Constraints without versions do not make sense: the role of a constraint
_is_ to influence dependency resolution by providing at least a version.
This should have been caught by Gradle and should in any case not trigger
a parsing bug (see https://github.com/gradle/gradle/issues/20182) but
that's how it is.

### DIFF
--- a/sql-bom/build.gradle
+++ b/sql-bom/build.gradle
@@ -9,6 +9,8 @@ micronautBom {
 }
 
 dependencies {
+    api platform("org.jdbi:jdbi3-bom:$jdbiVersion")
+    
     constraints {
 
         // vertx client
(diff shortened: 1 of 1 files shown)


timyates (approved):
Cool!  So the issue was the

```
        api "org.jdbi:jdbi3-core"
```

In the original PR?

Do we not require this?  Or does it get merged in automatically because of the platform dependency?

melix:
> Do we not require this? Or does it get merged in automatically because of the platform dependency?

That's correct. A user importing our BOM would _also_ get the `jdbi3-bom` transitively because of the `platform` dependency.

PR #592, opened 2022-04-27
Bump JDBC driver library versions

chadlwilson:
Similar to #530. Seems dependabot can't help us with these as it doesn't understand Gradle enough to know how to resolve the `gradle.properties`

- Most are minor bumps (in particular resolves a couple of CVEs in the current Postgres and Mysql driver versions) 
- MSSQL bumped to `9.4.1` ([changelog](https://docs.microsoft.com/en-us/sql/connect/jdbc/release-notes-for-the-jdbc-driver?view=sql-server-ver15#94)) 10.x has some breaking changes so probably needs to be done later
- Other minor versions are being bumped at #575 but as there are major upgrades there as well, it's causing failures.


chadlwilson:
Static analysis is failing due to a Sonarqube problem: `Unable to load component class org.sonar.scanner.scan.filesystem.ProjectFileIndexer` - don't think I caused this.

alvarosanchez:
No, the Sonar error is because GitHub won't expose the Sonar secret to your forked repo.

chadlwilson:
So does that just mean PRs from contributors should expect these build failures? Doesn't seem ideal if so?

alvarosanchez:
It's far from ideal, yes. But unfortunately, there's no way around it that we know. If you have any suggestion, we're open to change it.

In the meantime I've opened #595

PR #643, opened 2022-05-15
Use getHost instead of getContainerIpAddress

eddumelendez:
getContainerIpAddress is deprecated


graemerocher:
Thanks for the contribution!

PR #652, opened 2022-05-21
Add ability to configure ConnectionProvider for jOOQ

morki:
There was no way to configure custom `ConnectionProvider` so this PR adds one. 

The use case for this is to configure custom `ConnectionProvider` that sets authentication variables in every transaction for logging using trigger.


morki:
The build is failing with unrelated stuff, can someone please take a look?

dstepanov:
Follow that link to add a description of the change.

morki:
Thank you for your help @dstepanov, now it is ready for review.

PR #665, opened 2022-06-10
Support jOOQ commercial editions

morki:
Those types are available only in commercial jOOQ editions (oracle, sqlserver) and are needed for native image to compile.


dstepanov:
Thanks, @morki!

Issue #705, opened 2022-07-29
GraalVM 22.2 fails with NoClassDefFoundError: Could not initialize class org.mariadb.jdbc.internal.protocol.MasterProtocol


sdelamo:
Steps to reproduce: 

- [Clone repository `micronaut-graal-tests/micronaut-data-jdbc-graal` branch `3.6.x_mariadb`](https://github.com/micronaut-graal-tests/micronaut-data-jdbc-graal/tree/3.6.x_mariadb). 

Start maria locally via Docker. The app expects a `pets` database to exists and credentials: 

https://github.com/micronaut-graal-tests/micronaut-data-jdbc-graal/blob/3.6.x_mariadb/src/main/resources/application.yml#L9-L12

- Use GraalVM 22.2 `sdk use java 22.2.r11-grl`
- Build Native Image with ` ./build-native-image.sh`
- Execute Native Image `./data-jdbc-mariadb`

```   
 __  __ _                                  _   
|  \/  (_) ___ _ __ ___  _ __   __ _ _   _| |_ 
| |\/| | |/ __| '__/ _ \| '_ \ / _` | | | | __|
| |  | | | (__| | | (_) | | | | (_| | |_| | |_ 
|_|  |_|_|\___|_|  \___/|_| |_|\__,_|\__,_|\__|
  Micronaut (v3.6.0-SNAPSHOT)

11:23:53.579 [main] INFO  com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Starting...
11:23:53.580 [main] ERROR io.micronaut.runtime.Micronaut - Error starting Micronaut server: Bean definition [javax.sql.DataSource] could not be loaded: Error instantiating bean of type  [javax.sql.DataSource]

Message: Could not initialize class org.mariadb.jdbc.internal.protocol.MasterProtocol
Path Taken: DataSource.dataSource(DatasourceConfiguration datasourceConfiguration)
io.micronaut.context.exceptions.BeanInstantiationException: Bean definition [javax.sql.DataSource] could not be loaded: Error instantiating bean of type  [javax.sql.DataSource]

Message: Could not initialize class org.mariadb.jdbc.internal.protocol.MasterProtocol
Path Taken: DataSource.dataSource(DatasourceConfiguration datasourceConfiguration)
	at io.micronaut.context.DefaultBeanContext.initializeContext(DefaultBeanContext.java:1915)
[...]

PR #710, opened 2022-08-01
Add `java.sql.rowset` module for MariaDB connector.

fniephaus:
See https://github.com/micronaut-projects/micronaut-sql/issues/705


fniephaus:
You may want to add a comment that this is only needed for the MariaDB connector.

PR #725, opened 2022-09-19
ci: version to 5.0.0 Micronaut 4.0.0-SNAPSHOT githubCoreBranch=4.0.x remove deprecations

wetted:
I think this is right...

I had to disable the Graal tests and one r2dbc test, as I suspect the version of micronaut-data is not compatible.

Here: https://github.com/micronaut-projects/micronaut-sql/pull/725/files#diff-e7f94e7dee66c2740bb2526262d6860c09da1e9add7df9c524a37f18d6ffab35R33-R49

And here: https://github.com/micronaut-projects/micronaut-sql/pull/725/files#diff-e474ac6b4e6334a4416f9b0a1d4f66ce29aa1ea7a8ab5e84a353894529d1db28R20-R23

But we cannot fix micronaut-data until we have a version of micronaut-sql that works

I raised an issue to re-enable these

https://github.com/micronaut-projects/micronaut-sql/issues/751


sdelamo:
@wetted ping me once this is ready for review.

sdelamo (approved):

timyates:
This requires:

- https://github.com/micronaut-projects/micronaut-core/pull/8319
- Micronaut Data for 4.0.0-SNAPSHOT

before all the tests will pass

timyates:
Also we need this in core https://github.com/micronaut-projects/micronaut-core/pull/8327 (conversion for Properties)

timyates:
Think I found another core error... investigating

timyates:
Ok this is now blocked by:

1. We need a new version of Micronaut data that has a replacement for DefaultConversionService
2. UCP is failing to compile due to https://github.com/micronaut-projects/micronaut-core/issues/8410

timyates:
All we need now is a snapshot release of micronaut-data compiled against 4.0.0

timyates:
@wetted Can you give this a review before we merge it in?

timyates:
Ok, we have a chicken-egg situation here as this uses micronaut-data, and micronaut-data uses this

I've disabled the tests that I think are failing due to micronaut-data and it's dependencies to get a green build, and raised an issue to address them once we have a new version of micronaut-data

.github/workflows/central-sync.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         uses: actions/setup-java@v3
         with:
           distribution: 'adopt'
```
wetted, line 25: I think we want to use `temurin` now

wetted (commented):
Looks good to me. I think I would just change the java dists to `temurin` in the github workflows since we're starting to do that elsewhere. Thanks for helping out on this.

.github/workflows/central-sync.yml
timyates: Good catch 👍

PR #776, opened 2023-01-03
Update Oracle libs + snapshots

dstepanov:


graemerocher:
tests failing 🤔

dstepanov:
Not sure why it wants:
```
Execution failed for task ':sql-bom:checkVersionCatalogCompatibility'.
> Version catalogs are not compatible:
  The following libraries were present in the baseline version but missing from this catalog:
    - ojdbc8
```

Going to add `ojdbc8`

graemerocher:
would be a breaking change for users if that was removed I imagine

PR #781, opened 2023-01-18
Support for Hibernate6 - phase2

radovanradic:
Creating draft PR since there are some blocking issues so maybe someone could have a look.

1. EntityManager.persist(entity) does not trigger validation (not sure what is diff here so validation is not working)
2. Not able to register custom ByteCodeProvider that worked fine with hibernate5. Looks like someone also had this issue https://hibernate.atlassian.net/browse/HHH-13242 and it's not resolved yet (although marked as fixed in their JIRA)


hibernate6-jpa/src/test/groovy/io/micronaut/configuration/hibernate6/jpa/JpaSetupSpec.groovy
```diff
@@ -0,0 +1,229 @@
...
+        then:
+        entityManagerFactory != null
+
+        when:
+        EntityManager em = entityManagerFactory.createEntityManager()
+        def tx = em.getTransaction()
+        tx.begin()
+        em.persist(new Book(title: ""))
+        em.flush()
+        // TODO: persist/flush don't throw validation exception so do it manually
```
radovanradic, line 70: Validation not triggered and tests expect it and expect ConstraintValidationException. Any idea what is changed? I tested for javax.validation.Validator and there is `MicronautHibernateValidator` but it's not being called, not sure why.

hibernate6-jpa/src/main/java/org/hibernate/metamodel/internal/EntityRepresentationStrategyPojoStandard.java
```diff
@@ -0,0 +1,441 @@
...
+				mapsIdRepresentationStrategy = null;
+			}
+		}
+		else {
+			mapsIdRepresentationStrategy = null;
+			identifierPropertyName = identifierProperty.getName();
+			identifierPropertyAccess = makePropertyAccess( identifierProperty );
+		}
+
+		final BytecodeProvider bytecodeProvider = creationContext.getBootstrapContext().getServiceRegistry().getService(BytecodeProvider.class);
```
radovanradic, line 149: This is class copied from Hibernate and changed this line in order to use bytecode provider that we registered. Hibernate 6 uses `Environment.getBytecodeProvider();` and seems like there is no way to use custom. There is JIRA issue for this and I asked also in Hibernate ORM discussions about it.

hibernate6-jpa/src/test/groovy/io/micronaut/configuration/hibernate6/jpa/JpaSetupSpec.groovy
dstepanov: Because it's `javax` validator, now we have `jakarta`. I think this is expected.

radovanradic: Yes, `BeanValidationIntegrator` checks for existence of given class and does not find it 
```
org.hibernate.boot.registry.classloading.spi.ClassLoadingException: Unable to load class [jakarta.validation.ConstraintViolation]
```
So, what do we do with this? Are we going to get jakarta validation api through micronaut-validator future version? Can we release this without validator (in case we fix proxy factory issue)

hibernate6-jpa/src/main/java/org/hibernate/metamodel/internal/EntityRepresentationStrategyPojoStandard.java
radovanradic: Removed, no longer needed since we don't support compile time proxy in hibernate6

hibernate6-jpa/src/test/groovy/io/micronaut/configuration/hibernate6/jpa/JpaSetupSpec.groovy
radovanradic: Disabled test with validation and added couple of TODOs so we can add validation back with jakarta api.

hibernate6-jpa/build.gradle
```diff
@@ -16,12 +15,12 @@ dependencies {
...
     api(mn.micronaut.validation)
-    // TODO: Later will include mnData.micronaut.data.tx.hibernate6 after micronaut-data supports Hibernate6
-    //api(mnData.micronaut.data.tx.hibernate6)
+    api(mnData.micronaut.data.tx.hibernate6) {
+        exclude group: 'io.micronaut.sql'
+    }
     api(mn.micronaut.aop)
     api projects.jdbc
     api(mn.micronaut.runtime)
-    api(mn.micronaut.validation)
```
radovanradic, line 24: This was duplicated, same as line 43

hibernate6-jpa/src/main/java/io/micronaut/configuration/hibernate6/jpa/JpaConfiguration.java
```diff
@@ -60,7 +60,6 @@ public class JpaConfiguration {
     private List<String> mappingResources = new ArrayList<>();
     private EntityScanConfiguration entityScanConfiguration;
 
-    private boolean compileTimeHibernateProxies;
```
radovanradic, line 63: Not supported in hibernate6, no way to set custom byte code provider so dropped this for now.

radovanradic:
This is ready for review, for now no support for compile time proxies and validation

PR #801, opened 2023-02-07
build: update to Micronaut Framework 3.8.3

sdelamo:


sdelamo:
@radovanradic @dstepanov could you check this PR out? Upgrading to 3.8.3 breaks the build. `example.jooq.reactive.PostgresAppSpec` times out. https://ge.micronaut.io/s/fo5kscj7zwieo

radovanradic:
I looked little bit at this, sadly don't know enough what is going on behind the scenes, but seems like this test is failing 
```
when:
            def resultsList = Flux.range(1, 1000)
                    .flatMap { client.retrieve(HttpRequest.GET("/pets"), List) }
                    .collectList()
                    .block()
```
with 
```
Read Timeout
io.micronaut.http.client.exceptions.ReadTimeoutException: Read Timeout
```
as soon as 1000 is changed with 10 the test passes. When I increase it to 11 it fails again. So, seems like HttpClient cannot handle more than 10 requests, maybe not releasing handlers or no idea what could be the issue. Maybe Jonas would know more what could cause this when Micronaut is changed form 3.5.7 to 3.8.3

yawkat:
it does not look http-related, all the requests go through. i think it's a connection pool leak on the sql side. if you set a pool acquisition timeout it fails:

```patch
Index: tests/common-tests/src/main/groovy/example/sync/AbstractR2DBCContainerAppSpec.groovy
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/tests/common-tests/src/main/groovy/example/sync/AbstractR2DBCContainerAppSpec.groovy b/tests/common-tests/src/main/groovy/example/sync/AbstractR2DBCContainerAppSpec.groovy
--- a/tests/common-tests/src/main/groovy/example/sync/AbstractR2DBCContainerAppSpec.groovy	(revision abedadbb79b424af7ce404d4f0369bea8a23209a)
+++ b/tests/common-tests/src/main/groovy/example/sync/AbstractR2DBCContainerAppSpec.groovy	(date 1675945377168)
@@ -54,7 +54,7 @@
     String getR2dbUrlSuffix(String driverName, JdbcDatabaseContainer container) {
         switch (driverName) {
             case "postgresql":
-                return "${container.getHost()}:${container.getFirstMappedPort()}/${container.getDatabaseName()}?options=statement_timeout=5s"
+                return "${container.getHost()}:${container.getFirstMappedPort()}/${container.getDatabaseName()}?options=statement_timeout=5s&maxAcquireTime=PT1S"
             case "h2":
                 return "/testdb"
             case "sqlserver":
```

radovanradic:
Thanks @yawkat , it seems you are right. Connections are not being released and since pool max size is 10 that is why it fails when number of calls is >= 11. There have been some changes in micronaut data r2dbc transaction handling in 3.5.0 and later. This happens also in 4.0.0 but the test was disabled so we missed it.

radovanradic:
Ok, I think I fixed this issue in this commit https://github.com/micronaut-projects/micronaut-sql/pull/801/commits/5dab85d07dd93619942efcadf804034ba72bdfa9

PR #816, opened 2023-02-21
Refactored native image tests

msupic:
Refactored native image tests to be executed by the nativeTest task from NBT plugin and to use Test Resources plugin


.github/workflows/graalvm.yml
```diff
@@ -54,12 +54,7 @@ jobs:
       - name: Build with Gradle
         id: gradle
         run: |
-          if ./gradlew tasks --no-daemon --all | grep -w "testNativeImage"
-          then
-            ./gradlew check testNativeImage --continue --no-daemon
-          else
-            ./gradlew check --continue --no-daemon
-          fi
+          ./gradlew nativeTest --continue --no-daemon
```
graemerocher, line 57: rather than modifying this template which will just get overridden with the next project workflow sync can you instead use the approach like here:

https://github.com/micronaut-projects/micronaut-security/blob/686e2e509156057cd45226c6ea3423c9b7d85312/test-suite-graal/build.gradle#L22

graemerocher (changes requested):

.github/workflows/graalvm.yml
msupic: Yes, I just wanted to check whether everything is working as expected using the `nativeTest` task. Anyway if I use
```
tasks.named("check") { task ->
    def graal = ["jvmci.Compiler", "java.vendor.version", "java.vendor"].any {
        println "$it ${System.getProperty(it)}"
        System.getProperty(it)?.toLowerCase(Locale.ENGLISH)?.contains("graal")
    }
    if (graal) {
        task.dependsOn("nativeTest")
    }
}
```
the native tests will be duplicated (executed once by old `testNativeImage` task and once by `nativeTest` task) because of the following code in the template (https://github.com/micronaut-projects/micronaut-project-template/blob/master/.github/workflows/graalvm.yml)
```
          if ./gradlew tasks --no-daemon --all | grep -w "testNativeImage"
          then
            ./gradlew check testNativeImage --continue --no-daemon
          else
            ./gradlew check --continue --no-daemon
          fi
```
Since that template is used in all projects, I can't remove `testNativeImage` from `./gradlew check testNativeImage --continue --no-daemon`.  So I need to remove `testNativeImage` tasks from micronaut-sql module. To do that I need to upgrade micronaut-gradle-plugin to 4.0.0-SNAPSHOT (which removed `testNativeImage` task) for micronaut-sql module and currently I am trying to figure out how to do that.

graemerocher: probably easier to just disable the `testNativeImage` task:

```
testNativeImage.enabled = false
```

graemerocher (approved):

PR #822, opened 2023-03-01
Require Java 11 for micronaut-jooq and patch update jooq

### LINKED ISSUE #821: Jooq 3.15.x requires Java 11
In v4.5.0 we [downgraded Jooq to 3.15.10](https://github.com/micronaut-projects/micronaut-sql/pull/678) in an attempt to get a version with Java 8 support.

Unfortunately, Java 7 was dropped from v3.15.x onwards

We need to downgrade to 3.14.16 to support Java 8

(Search for "Drop Java 6/7 support for Enterprise Edition, require Java 11 in OSS Edition" [here](https://www.jooq.org/notes/?version=3.15#:~:text=Drop%20Java%206/7%20support%20for%20Enterprise%20Edition%2C%20require%20Java%2011%20in%20OSS%20Edition))

timyates:
In 4.5.0 we upgraded Jooq to get R2DBC support.

As Jooq 3.15.x requires Java 11, the Jooq module was marked as Java 11 only (in v4.5.x)

This restriction was removed in 4.6.x, however jOOQ still required Java 11.

Normally, we would downgrade jOOQ to maintain Java 8 support, however R2DBC support only exists in 3.15+ so we cannot do this.

This PR:

- re-introduces the java 11 compile target for the jOOQ module
- adds a note to the documentation to say that the jooq module requires Java 11+
- Updates jOOQ to the latest patch release of 3.15 (at time of writing)

Closes #821

We should merge this up for Micronaut 3.8.7

### DIFF
--- a/.github/workflows/graalvm.yml
+++ b/.github/workflows/graalvm.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         java: ['11', '17']
-        graalvm: ['latest', 'dev']
+        graalvm: ['latest']
     steps:
        # https://github.com/actions/virtual-environments/issues/709
       - name: Free disk space
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -9,7 +9,7 @@ micronaut-gradle-plugin = "3.6.0"
 # Frameworks
 
 managed-vertx = "4.3.3"
-managed-jooq = "3.15.11"
+managed-jooq = "3.15.12"
 managed-hibernate = "5.6.11.Final"
 managed-hibernate-reactive = "1.1.7.Final"
 managed-jasync = "2.0.8"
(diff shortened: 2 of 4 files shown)


sdelamo (approved):

sdelamo:
@timyates can you create an issue in starter add it to 3.8.7 board to restrict jooq feature to java 11 or 17. We should throw an exception in starter if you select Java 8

melix (approved):
Since it's a breaking change, should it be `5.0.0`?

sdelamo:
> Since it's a breaking change, should it be `5.0.0`?

I think the issue is it is already broken. We release a module which does not work with java 8.

timyates:
> Since it's a breaking change, should it be `5.0.0`?

It's broken now (was broken in 4.5.0 last July), this is just documenting the breakage as we cannot roll it back (due to R2DBC)

timyates:
@sdelamo [Issue](https://github.com/micronaut-projects/micronaut-starter/issues/1637) and [PR](https://github.com/micronaut-projects/micronaut-starter/pull/1638) raised for starter for 3.8.7

PR #831, opened 2023-03-20
Additional updates for micronaut-sql

radovanradic:
@graemerocher This branch is created from your branch https://github.com/micronaut-projects/micronaut-sql/tree/build-update-and-jakarta-validation and added more needed changes. Some tests are disabled now. If it's ok, we can merge this to your branch and continue there. 
If having few disabled tests are ok, then next step would be to remove hibernate5 completely.


hibernate6-jpa/src/main/java/io/micronaut/configuration/hibernate6/jpa/graal/HibernateSubstitutions.java
```diff
@@ -138,6 +143,19 @@ final class Hql {
...
+// Disable Runtime Byte Code Enhancement
+@TargetClass(className = "org.hibernate.cfg.Environment")
+@TypeHint(
+    value = {EventType.class, EnumType.class},
+    accessType = {AccessType.ALL_DECLARED_FIELDS, AccessType.ALL_DECLARED_METHODS, AccessType.ALL_DECLARED_CONSTRUCTORS}
+)
+final class EnvironmentSubs {
+    @Substitute
+    public static BytecodeProvider buildBytecodeProvider(Properties properties) {
+        return new IntrospectedHibernateBytecodeProvider();
```
radovanradic, line 155: Added back since now it is possible to inject custom bytecode provider (hibernate 6.2).

hibernate6-jpa/src/main/java/io/micronaut/configuration/hibernate6/jpa/proxy/IntrospectedHibernateBytecodeProvider.java
```diff
@@ -0,0 +1,65 @@
...
+    public ReflectionOptimizer getReflectionOptimizer(Class clazz, String[] getterNames, String[] setterNames, Class[] types) {
+        // This is deprecated and no longer used
+        return null;
+    }
+
+    @Override
+    public ReflectionOptimizer getReflectionOptimizer(Class<?> clazz, Map<String, PropertyAccess> propertyAccessMap) {
+        // Prev implementation doesn't return accurate optimizer and bean properties
+        // So some tests are failing. Returning null fixes some failing tests for now
+        return null;
```
radovanradic, line 57: It appears this never worked in prev micronaut-sql versions as method was this
```
@Override
                    public Object[] getPropertyValues(Object object) {
                        Object[] values = new Object[beanProperties.length];
                        for (int i = 0; i < beanProperties.length; i++) {
                            BeanProperty beanProperty = beanProperties[i];
                            values[i] = beanProperty.get(i);
                        }
                        return values;
                    }
```
This line `values[i] = beanProperty.get(i);` should be `values[i] = beanProperty.get(object );` or otherwise it would fail. But, when I change it then still getting errors since this doesn't return matching field names with what Hibernate expects. This for instance returns 3 fields in one of our tests, hibernate expects 2 and in different order. Since it didn't work before, I guess we can ignore it. The reason it was not worked is flag `hibernate.bytecode.use_reflection_optimizer` in hibernate that was turned off by default in hibernate5 and now is enabled in hibernate6. And we would get issue reported if someone had it enabled.

tests/common-tests/src/main/java/example/sync/AbstractApp.java
```diff
@@ -95,6 +91,7 @@ void shouldFetchOwnerByName() {
 
     @Test
     @Order(4)
+    @Disabled("Getting Flux (findAll) is failing using hibernate-reactive 2.0")
```
radovanradic, line 94: These tests are temp disabled for hibernate-reactive

dstepanov (approved):

tests/common-tests/src/main/java/example/sync/AbstractApp.java
radovanradic: This is fixed now, hibernate-reactive folks suggested improved code that works ok now.

PR #881, opened 2023-05-05
Disable JPA via configuration

sdelamo:
This would be ease for example project generation in starter. Right now we generate an application without any entity. Thus the application tests will fail until the user creates a `@Entity` and defines `jpa.default.entity-scan.packages`


sdelamo:
failure Gradle scan https://ge.micronaut.io/s/in4pzerinyyl2

graemerocher:
@sdelamo please rebase

sdelamo:
> @sdelamo please rebase

done

graemerocher:
The name of the configuration option is not very consistent with the existing no (it uses `micronaut.hibernate.jpa` instead of just `jpa`) and also doesn't allow disabling individual JPA configurations.

WDYT @dstepanov ?

dstepanov:
Would it be better to create a simple entity in the starter?

sdelamo:
@dstepanov can we merge this?

graemerocher:
my comment was never addressed, why the inconsistent name?

graemerocher (approved):

dstepanov (approved):

PR #885, opened 2023-05-11
Fixed micronaut-serde conflict issue

msupic:


graemerocher:
test failures?

msupic:
I thought that solving of version conflict issue might solve the following issue
`Cannot invoke "io.micronaut.serde.Deserializer.deserializeNullable(io.micronaut.serde.Decoder, io.micronaut.serde.Deserializer$DecoderContext, io.micronaut.core.type.Argument)" because "this.deserializer" is null`
but it looks like there is something else. I am investigating that.

msupic:
Closing since done in https://github.com/micronaut-projects/micronaut-sql/pull/863/

PR #971, opened 2023-06-13
Don't use deprecated class from Micronaut Data

dstepanov:


radovanradic:
Added more changes to fix compilation error after micronaut-data M9 was added. Currently jooq r2dbc tests are failing. Seems like connection is not being released properly and these tests run 1000 calls in parallel. After 10th connection is established application is blocked because new connection cannot be obtained (10 is max).  @dstepanov Looks like Flux.usingWhen asyncCleanup handler is not being called immediately after work has been done, but couldn't really figure out what is going on here.

PR #998, opened 2023-06-29
Build: update binary compatibility check baseline

wetted:


buildSrc/src/main/groovy/io.micronaut.build.internal.sql-module.gradle
```diff
@@ -5,7 +5,7 @@ plugins {
 
 micronautBuild {
     binaryCompatibility {
-        baselineVersion = "5.0.0-M10"
+        baselineVersion = "5.0.0-M12"
```
sdelamo, line 8: is the baselineVersion required or can we get rid of it?

sdelamo (approved):

Issue #1266, opened 2024-02-12
Multiple possible bean candidates found: [`DefaultInternalConstraintValidatorFactory`, `DefaultConstraintValidatorFactory`]


tcrespog:
### Expected Behavior

No exception on starting up a default Micronaut Data + Hibernate JPA app.

### Actual Behaviour

The application fails to start with the following error:
```
ERROR io.micronaut.runtime.Micronaut - Error starting Micronaut server: Bean definition [org.hibernate.SessionFactory] could not be loaded: Error instantiating bean of type  [io.micronaut.configuration.hibernate.jpa.conf.sessionfactory.configure.internal.ValidatorFactoryConfigurer]

Message: Multiple possible bean candidates found: [DefaultInternalConstraintValidatorFactory, DefaultConstraintValidatorFactory]
Path Taken: SessionFactoryPerDataSourceFactory.buildHibernateSessionFactoryBuilder(SessionFactoryBuilder sessionFactoryBuilder) --> new SessionFactoryPerDataSourceFactory(Environment environment,[List configures],StandardServiceRegistryBuilderCreator serviceRegistryBuilderSupplier,List standardServiceRegistryBuilderConfigurers,JpaConfiguration jpaConfiguration,ApplicationContext applicationContext,Integrator integrator) --> new ValidatorFactoryConfigurer([ValidatorFactory validatorFactory])
io.micronaut.context.exceptions.BeanInstantiationException: Bean definition [org.hibernate.SessionFactory] could not be loaded: Error instantiating bean of type  [io.micronaut.configuration.hibernate.jpa.conf.sessionfactory.configure.internal.ValidatorFactoryConfigurer]
```

### Steps To Reproduce

1. Create a Micronaut Data + Hibernate JPA project using Micronaut 4.3.1: `mn create-app -b gradle -l groovy -t spock --features=h2,jdbc-tomcat,data-jpa,hibernate-jpa,hibernate-validator mn-hibernate-playground`.
2. Run the app: `./gradlew run`.
3. The app fails to start.

### Environment Information

- Ubuntu 22.04
- Java 17

### Version

4.3.1

tcrespog:
Sorry, but the error seems to keep happening in Micronaut 4.3.2 (same steps to reproduce).

sdelamo:
A workaround to force the usage of the Micronaut hibernate validator `ConstraintValidatorFactory` is to create such as class:

``java
package mn.hibernate.playground;

[...]

PR #1267, opened 2024-02-12
fix: don’t leakage Micronaut Validation

sdelamo:
Close: https://github.com/micronaut-projects/micronaut-sql/issues/1266

### DIFF
--- a/hibernate-jpa/build.gradle
+++ b/hibernate-jpa/build.gradle
@@ -7,7 +7,7 @@ dependencies {
 
     api(libs.managed.hibernate.core)
     api(libs.managed.jakarta.transaction.api)
-    implementation(mnValidation.validation) // jakarta.validation:jakarta.validation-api
+    compileOnly(mnValidation.validation) // jakarta.validation:jakarta.validation-api
     compileOnly(mnData.micronaut.data.tx.hibernate)
     testImplementation(mnData.micronaut.data.tx.hibernate)
     api(mn.micronaut.aop)
--- a/hibernate-jpa/build.gradle
+++ b/hibernate-jpa/build.gradle
@@ -28,7 +28,7 @@ dependencies {
     testImplementation(mnMicrometer.micronaut.micrometer.core)
     testImplementation(libs.managed.hibernate.micrometer)
 
-    testImplementation(mnValidation.micronaut.validation)
+    testImplementation(mnHibernateValidator.micronaut.hibernate.validator)
 
     testRuntimeOnly projects.micronautJdbcTomcat
     testRuntimeOnly(libs.managed.h2)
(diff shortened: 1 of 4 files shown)


hibernate-jpa/build.gradle
```diff
@@ -7,7 +7,7 @@ dependencies {
 
     api(libs.managed.hibernate.core)
     api(libs.managed.jakarta.transaction.api)
-    api(mnValidation.micronaut.validation)
+    implementation(mnValidation.validation) // jakarta.validation:jakarta.validation-api
```
dstepanov, line 10: Why do we need to include the validation anyway?

radovanradic (approved):

hibernate-jpa/build.gradle
graemerocher: I don't think this will fix the issue

sdelamo: @dstepanov We need the dependency to `jakarta.validation:jakarta.validation-api` because of [ValidatorFactorySettingSupplier](https://github.com/micronaut-projects/micronaut-sql/blob/master/hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/conf/settings/internal/ValidatorFactorySettingSupplier.java). 

@graemerocher 
> I don't think this will fix the issue

In this PR for `hibernate-jpa`build I included: 

`testImplementation(mnHibernateValidator.micronaut.hibernate.validator)`

It seems to get rid of the `Multiple possible bean candidates found: [DefaultInternalConstraintValidatorFactory, DefaultConstraintValidatorFactory]` error.

dstepanov: Put it compile only

sdelamo: done https://github.com/micronaut-projects/micronaut-sql/pull/1267/commits/09609a2793f82bf6ea335a4d4e29ab72765c7df9

dstepanov (approved):

graemerocher (approved):

graemerocher:
Regarding this, it is a patch/workaround to the actual issue. IMO the Hibernate Validator module should be updated to allow it to live side by side with the Micronaut validation module

dstepanov:
The changes in Micronaut Serialization actually broke it, but I think the combination of two shouldn’t be allowed anyway

radovanradic:
@sdelamo Is this going to be reverted maybe after this PR https://github.com/micronaut-projects/micronaut-hibernate-validator/pull/370 ?
Asking because it seems it caused issue in micronaut-data https://github.com/micronaut-projects/micronaut-data/issues/2808
Validation is removed and then Hibernate behaves differently, it validates nullability before listeners are fired and created date populated. Talked about nullability validation here https://stackoverflow.com/questions/66524227/columnnullable-false-validating-the-null-check-at-application-level

sdelamo:
> We might need to make remove that second bean from Micronaut Validation, like it was before

@dstepanov I don't follow what you suggest.

dstepanov:
@sdelamo https://github.com/micronaut-projects/micronaut-validation/commit/be706d04830f4bc7dccd113d9b826c1ebd26119f

PR #1268, opened 2024-02-12
Revert ojdbc version 23.3.0.23.09 back to 21.11.0.0

radovanradic:
ucp 23.3.0.23.9 library doesn't work in a native image. Until fixed we should stick to 21.11.0.0.

I have added jdbc-ucp native test and then it will fail as renovate creates PR to upgrade to 23.x until the issue is fixed.

Cc: @msupic , @graemerocher

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -31,7 +31,7 @@ managed-jdbi = "3.44.1"
 
 # JDBC Drivers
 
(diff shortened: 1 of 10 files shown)


graemerocher (approved):

sdelamo (approved):
I think we should make Micronaut Oracle cloud expose as a  managed version 
ojdbc and then consume it Micronaut SQL so that we don't have two versions defined.

PR #1276, opened 2024-02-16
Define Oracle ojdbc version using BOM exposed by micronaut-oracle-cloud.

### LINKED ISSUE #1274: don't define ojdbc version instead use the one exposed by oracle cloud
### Issue description

see: https://github.com/micronaut-projects/micronaut-oracle-cloud/pull/804

wetted:
closes #1274

A few things of note (questions):

- I think `micronaut-oracle-cloud` needs to also expose the Oracle `ucp` lib with `managed-oracle-jdbc-ucp = { module = 'com.oracle.database.jdbc:ucp', version.ref = 'managed-ojdbc'  }`.  (re: https://github.com/micronaut-projects/micronaut-oracle-cloud/pull/804). Otherwise, it forces us to use a platform dependency in the module builds
  - maybe `com.oracle.database.observability` libs too?
- Removing the managed ojdbc libs from the `micronaut-sql` version catalog is a breaking change so I bumped to next major
- I'm not so sure I handled the gradle build for `:sql-bom` correctly. Guidance appreciated if I didn't.


gradle.properties
```diff
@@ -1,4 +1,4 @@
-projectVersion=5.5.2-SNAPSHOT
+projectVersion=6.0.0-SNAPSHOT
```
sdelamo, line 1: Did you change it to the next major because we no longer exposed these managed versions?

sdelamo (approved):

gradle.properties
wetted: > Did you change it to the next major because we no longer exposed these managed versions?

Yes, see my note in the description. I am erring on the side of caution here.

sdelamo:
> Moving this to draft pending: [micronaut-projects/micronaut-oracle-cloud#821](https://github.com/micronaut-projects/micronaut-oracle-cloud/pull/821)
> 
> I also want to document in breaking changes section.

I don't think we need #821. We can just use the bom as you were doing. The goal is just to avoid define a version number twice . Please add breaking changes  documentation.

wetted:
> I don't think we need #821. We can just use the bom as you were doing. The goal is just to avoid define a version number twice . Please add breaking changes documentation.

OK. I added the docs earlier for your review.

melix (commented):
I'm not sure if that is how we should do it. Either micronaut-sql should use micronaut-oracle-cloud (this PR) or the other way around. However, from my point of view, it should rather be the other way around, because I don't see why I would need a "cloud" module just to get a JDBC driver version.

melix:
With this is would basically mean create one module per provider. `micronaut-postgresql`, etc... It makes more sense to me that the SQL module defines what drivers it is compatible with.

wetted:
I agree with @melix and wondered about it before. 

Also, `micronaut-oracle-cloud` already imports the `micronaut-sql` catalog anyway, so getting the jdbc drivers from `micronaut-sql` makes sense to me. Doing it the other way around adds a circular dependency on module catalogs.

sdelamo:
> I agree with @melix and wondered about it before.
> 
> Also, `micronaut-oracle-cloud` already imports the `micronaut-sql` catalog anyway, so getting the jdbc drivers from `micronaut-sql` makes sense to me. Doing it the other way around adds a circular dependency on module catalogs.

If oracle cloud already was consuming sql we should not have made ojdbc dependency managed and consume the one from sql.

PR #1323, opened 2024-04-24
Allow disabling of datasources by excluding packages

timyates:
This allows the testing of non-db beans without requiring the connection pools to be fully configured.

Does not cover r2dbc as that is defined elsewhere.

https://github.com/micronaut-projects/micronaut-data/issues/2893

It is done this way (with package requirements) as there is no way currently to allow EachBean to skip factory creation based on a property of the bean (ie Toggleable#isEnabled).

### DIFF
--- a/jdbc-hikari/src/main/java/io/micronaut/configuration/jdbc/hikari/package-info.java
+++ b/jdbc-hikari/src/main/java/io/micronaut/configuration/jdbc/hikari/package-info.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 original authors
+ * Copyright 2017-2024 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


sdelamo (approved):

jdbc-hikari/src/main/java/io/micronaut/configuration/jdbc/hikari/package-info.java
```diff
@@ -18,8 +18,11 @@
  */
 @Configuration
 @Requires(classes = HikariDataSource.class)
```
dstepanov, line 20: Maybe add requires to BasicJdbcConfiguration and that should be enough for all implementations

timyates: I'll try 🤔

timyates: @dstepanov No, the `@Requires` doesn't seem to carry through...  In my tests with Hikari, we still end up with a `DatasourceConfiguration` and a `HikariUrlDataSource` bean

dstepanov (approved):

sdelamo (approved):

PR #1388, opened 2024-07-29
Change the way datasources are being disabled

radovanradic:
This is followup for https://github.com/micronaut-projects/micronaut-sql/pull/1323 and slight change how datasources can be disabled.
Using global `datasources.enabled` was problematic because actually datasource configuration named `enabled` was being created in the context and when there are test-resources then it tries to resolve properties for datasource named `enabled` and then throws the error. This approach is maybe a bit better because each datasource can be enabled or disabled.
Will need followup PR in micronaut-data to skip schema generation for disabled datasources.

### DIFF
--- a/jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
+++ b/jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
@@ -47,8 +47,7 @@ import java.util.Map;
  */
 @Context
 @EachProperty(value = BasicJdbcConfiguration.PREFIX, primary = "default")
-@Internal
-class DatasourceConfiguration extends BasicDataSource implements BasicJdbcConfiguration {
+public class DatasourceConfiguration extends BasicDataSource implements BasicJdbcConfiguration {
 
     private static final Logger LOG = LoggerFactory.getLogger(DatasourceConfiguration.class);
     private final CalculatedSettings calculatedSettings;
@@ -188,7 +187,7 @@ class DatasourceConfiguration extends BasicDataSource implements BasicJdbcConfig
      * @param enabled an indicator telling whether data source is enabled
      */
     @Internal
-    public void setEnabled(boolean enabled) {
+    void setEnabled(boolean enabled) {
         if (!enabled) {
             // This is the only way to disable this bean which is actual datasource
             // because dbcp doesn't have datasource factory like other datasource implementations
--- a/jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
+++ b/jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
@@ -19,6 +19,7 @@ import io.micronaut.context.annotation.Context;
 import io.micronaut.context.annotation.EachProperty;
 import io.micronaut.context.annotation.Parameter;
 import io.micronaut.context.annotation.Property;
+import io.micronaut.context.exceptions.DisabledBeanException;
 import io.micronaut.core.convert.format.MapFormat;
 import io.micronaut.core.naming.conventions.StringConvention;
 import io.micronaut.jdbc.BasicJdbcConfiguration;
@@ -50,6 +51,7 @@ public class DatasourceConfiguration extends BasicDataSource implements BasicJdb
     private static final Logger LOG = LoggerFactory.getLogger(DatasourceConfiguration.class);
     private final CalculatedSettings calculatedSettings;
     private final String name;
+    private boolean enabled = true;
 
     /**
      * Constructor.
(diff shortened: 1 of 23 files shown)


radovanradic:
The reported issue is https://github.com/micronaut-projects/micronaut-data/issues/3036

dstepanov:
Does it support disabling with the previous syntax?

radovanradic:
No, we would then have the same issue as reported in the ticket. Our datasource are defined with prefix `datasources.<ds-name>.` and if we have property `datasources.enabled.` then it will be treated as one of datasources. Disablement on package level worked expecting exactly `datasources.enabled` property to be false. We don't have it now.

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
```diff
@@ -176,4 +178,22 @@ public void setDataSourceProperties(@MapFormat(transformation = MapFormat.MapTra
...
+    @Override
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    /**
+     * Sets an indicator telling whether data source is enabled.
+     * @param enabled an indicator telling whether data source is enabled
+     */
+    public void setEnabled(boolean enabled) {
```
graemerocher: can we make this package private or annotate with `@Internal`? Since it always throws probably not ideal

radovanradic: I marked class with `@Internal` and made non `public`. Tested with micronaut-data and examples used to reproduce the original issues, seems to be working ok.

dstepanov:
> No, we would then have the same issue as reported in the ticket. Our datasource are defined with prefix `datasources.<ds-name>.` and if we have property `datasources.enabled.` then it will be treated as one of datasources. Disablement on package level worked expecting exactly `datasources.enabled` property to be false. We don't have it now.

You can probably implement some custom check like `AbstractHibernateCondition` in Micronaut Data to avoid the enable.

radovanradic:
> > No, we would then have the same issue as reported in the ticket. Our datasource are defined with prefix `datasources.<ds-name>.` and if we have property `datasources.enabled.` then it will be treated as one of datasources. Disablement on package level worked expecting exactly `datasources.enabled` property to be false. We don't have it now.
> 
> You can probably implement some custom check like `AbstractHibernateCondition` in Micronaut Data to avoid the enable.

Yes, I tried that first and it resolves configuration and disables it. But, it looks like because it is `@EachProperty`
```
@EachProperty(value = BasicJdbcConfiguration.PREFIX, primary = "default")
public class DatasourceConfiguration extends HikariConfig implements BasicJdbcConfiguration {
```
then if we have multiple datasources and we want some of them to be enabled, it won't work because of that condition all beans will be disabled.
But, I will try this approach one more time.

radovanradic:
> > No, we would then have the same issue as reported in the ticket. Our datasource are defined with prefix `datasources.<ds-name>.` and if we have property `datasources.enabled.` then it will be treated as one of datasources. Disablement on package level worked expecting exactly `datasources.enabled` property to be false. We don't have it now.
> 
> You can probably implement some custom check like `AbstractHibernateCondition` in Micronaut Data to avoid the enable.

Thanks for suggestion.
It worked for the `DataSourceFactory.dataSource` methods, only for DBCP had to stick with that workaround with `setEnabled(boolean)` that throws `DisabledBeanException` because there is no factory and configuration is actually data source.

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
```diff
@@ -45,7 +47,8 @@
  */
 @Context
 @EachProperty(value = BasicJdbcConfiguration.PREFIX, primary = "default")
-public class DatasourceConfiguration extends BasicDataSource implements BasicJdbcConfiguration {
+@Internal
+class DatasourceConfiguration extends BasicDataSource implements BasicJdbcConfiguration {
```
graemerocher: this is a breaking change

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
graemerocher: class can't be made non-public it would be a breaking change

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
radovanradic: I see, reverted and changed `setEnabled` method non public with docs about the behavior and intention of the method.

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
radovanradic: Yes, reverted it and updated method only so it works as needed.

sdelamo (approved):
@radovanradic it looks better to be able to disable them individually. Thanks for this PR. I  would like to merge this breaking change notice though https://github.com/micronaut-projects/micronaut-sql/pull/1399

dstepanov (approved):

PR #1487, opened 2024-11-12
allow disabling JPA per datasource

graemerocher:
Currently you can only disable JPA for all datasources, but you might want to use JPA for one datasource but not the other. This PR adds the ability to disable JPA for a particular datasource.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -45,7 +45,7 @@ jobs:
           fetch-depth: 0
 
       - name: "🔧 Setup GraalVM CE"
-        uses: graalvm/setup-graalvm@v1.2.3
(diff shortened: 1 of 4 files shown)


gradle/libs.versions.toml
```diff
@@ -37,7 +37,7 @@ managed-postgres-driver = "42.7.4"
 managed-mariadb-driver = "3.4.1"
 managed-mssql-driver = "12.8.1.jre11"
 managed-mysql-driver = "8.0.33"
-managed-mysql-connector-j = "8.4.0"
+managed-mysql-connector-j = "9.1.0"
```
radovanradic: Should this go into the patch?

radovanradic (approved):

PR #1545, opened 2025-02-17
Sql to Core 4.8.x

sdelamo:


radovanradic:
Looks like these tests are failing because of https://github.com/micronaut-projects/micronaut-sql/blob/6.1.x/jdbc/src/main/java/io/micronaut/jdbc/JdbcDataSourceEnabled.java added recently to support disabling datasources. Don't know if there was some improvement work in core 4.8.x, but looks like this condition is cached and only executed for one data source configuration although we have it declared here for each `DatasourceConfiguration`
```
@Context
    @EachBean(DatasourceConfiguration.class)
    @Requires(condition = JdbcDataSourceEnabled.class)
    public DataSource dataSource(DatasourceConfiguration datasourceConfiguration) {
        HikariUrlDataSource ds = new HikariUrlDataSource(datasourceConfiguration);
        addMeterRegistry(ds);
        dataSources.add(ds);
        return ds;
    }
```
Do we need different approach for requires condition now?

dstepanov:
There is a compilation issue

sdelamo:
This test fails

`./gradlew :micronaut-jdbc-hikari:test --tests "io.micronaut.configuration.jdbc.hikari.DatasourceConfigurationSpec"`

@dstepanov can you check it?

dstepanov:
@radovanradic Where is it cached?

radovanradic:
I just asked if there were some changes in core, when I debugged this test I noticed `@Requires(condition = JdbcDataSourceEnabled.class)` is called only once with core 4.8.x and in 4.7.x it is called for each data source configuration.

radovanradic:
This was fixed using core 4.8.5

PR #1551, opened 2025-02-28
Added configuration option to destroy UCP pool

radovanradic:


jdbc-ucp/src/main/java/io/micronaut/configuration/jdbc/ucp/OracleUcpConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+import static io.micronaut.configuration.jdbc.ucp.OracleUcpConfiguration.PREFIX;
+
+/**
+ * Configuration properties for Oracle Universal Connection Pooling (UCP).
+ *
+ * @param destroyOnReload Flag indicating whether to destroy connections on reload
+ */
+@ConfigurationProperties(PREFIX)
+@Requires(property = PREFIX)
+public record OracleUcpConfiguration(@Nullable Boolean destroyOnReload) {
```
graemerocher, line 31: does it make sense to combine this with https://github.com/micronaut-projects/micronaut-sql/blob/6.1.x/jdbc-ucp/src/main/java/io/micronaut/configuration/jdbc/ucp/UniversalConnectionPoolManagerConfiguration.java ?

radovanradic: I thought there might be other `oracle.ucp.*` properties to set and we can use this for that purpose and looks to me it's better to keep separated from the ucp-manager.

sdelamo: I rescheduled this to 4.9

graemerocher (approved):

Issue #1398, opened 2024-08-12
Expose Hikari DataSources in Beans to support password rotation


napsta32:
### Problem

Currently the DataSource's are encapsulated inside the `DatasourceFactory` class (see https://github.com/micronaut-projects/micronaut-sql/blob/5.7.x/jdbc-hikari/src/main/java/io/micronaut/configuration/jdbc/hikari/DatasourceFactory.java). This makes it impossible to access the datasource without reflection.

I would like to access the DataSource instances to rotate passwords without affecting the service availability.

### Feature idea

Ideally I would like to simply get the "DataSource" and change things directly in there:

```java
// I would use either DataSource.class or HikariDataSource.class
HikariDataSource dataSource = applicationContext.getBean(DataSource.class, Qualifiers.ofName("datasourceName"));
dataSource.setPassword(newPassword);
dataSource.getHikariPoolMXBean().softEvictConnections(); // Refresh connections
```

### Workaround (using reflection)

As an alternative I am working with the following workaround (using reflection):

```java
    private void updateDataSourcePassword() throws NoSuchFieldException, IllegalAccessException {
        DatasourceFactory datasourceFactory = applicationContext.getBean(DatasourceFactory.class);
        Field dataSourcesField = DatasourceFactory.class.getDeclaredField("dataSources");
        dataSourcesField.setAccessible(true);

        List<HikariDataSource> dataSources = (List<HikariDataSource>) dataSourcesField.get(datasourceFactory);
        for (HikariDataSource dataSource : dataSources) {
            dataSource.setPassword(newPassword);
            dataSource.getHikariPoolMXBean().softEvictConnections(); // Refresh connections
        }
    }
```

Please let me know if there is a better workaround.

### How does Spring work (probably the best solution)

This is something kindof supported by Spring:
* https://blog.jdriven.com/2021/06/configure-hikari-connection-pool-when-using-rds-iam/
* https://stackoverflow.com/questions/73793784/how-to-ensure-hikari-connections-are-spawn-with-latest-credentials

Basically you can override the `HikariUrlDataSource` class with any custom classname using the `application.properties`. This is something that is not supported either by Micronaut. It is quite a clean solution with the disadvantage that `getPassword` is called for each connection. It could be worked around using a memory cache or lazy variables.

graemerocher:
it is not impossible, all you need to do is define a bean that implements `BeanCreatedEventListener<DataSource>` and intercept each beans creation

[...]

PR #1599, opened 2025-05-08
Support datasource password change in runtime

radovanradic:
This PR adds support for changing datasource password without need to restart the app. It might address some reported issues like [this one](https://github.com/micronaut-projects/micronaut-sql/issues/1398).
The idea is that users publish event with datasource name and changed password so listeners would be able to update datasource passwords. Current connections created with old password are validated (tested with hikari and ucp) and only when connection expires and being recreated then new password will be used.
@sdelamo Suggested using bean refresh, but don't know how it would work - tried refreshing `DataSourceConfiguration` but that didn't have effect since `DataSource` beans are already created and injected with the old password and refreshing `DataSourceConfiguration` won't _rebuild_ existing datasources. Or maybe I was not aware how it could be done.


tests/hibernate6/hibernate6-reactive-mysql/src/test/java/example/hibernate6/reactive/MySQLApp.java
```diff
@@ -22,5 +22,6 @@
 @MicronautTest(transactional = false)
 @Property(name = "jpa.default.properties.hibernate.connection.db-type", value = "mysql")
 @Property(name = "jpa.default.reactive", value = "true")
+@Property(name = "test-resources.containers.mysql.image-name", value = "mysql:8.4.5")
```
radovanradic: Unrelated to this PR, `mysql:latest` docker image was failing to start in tests so needed to update it to 8.x

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceFactory.java
```diff
@@ -56,9 +65,23 @@ public DataSourcePoolMetadata<BasicDataSource> dbcpDataSourcePoolMetadata(
...
-        if (resolved instanceof BasicDataSource) {
-            dbcpDataSourcePoolMetadata = new DbcpDataSourcePoolMetadata((BasicDataSource) resolved);
+        if (resolved instanceof BasicDataSource basicDataSource) {
+            dbcpDataSourcePoolMetadata = new DbcpDataSourcePoolMetadata(basicDataSource);
         }
         return dbcpDataSourcePoolMetadata;
     }
+
+    @Override
+    public void onApplicationEvent(DataSourcePasswordChangedEvent event) {
```
graemerocher: I think instead of a new event you should extend `RefreshEventListener` and override `getObservedConfigurationPrefixes()` to get the data which will make it work in other contexts like the refresh endpoint and periodic monitoring.

For UCP you should call:

```
public void reconfigureDataSource(Properties configuration) throws SQLException {
```

For Hikari you should call:

```
hikariDataSource.getHikariConfigMXBean().setPassword("newPassword");
hikariDataSource.getHikariPoolMXBean().softEvictConnections();
```

For DBCP:

```
// set the properties
// and then call
pool.evict();
```

graemerocher (changes requested):

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceFactory.java
radovanradic: I wasn't sure if I am able to populate all `Properties` for UCP or can only set password to properties?
Btw, I think this is not needed, since existing connections will have old password and it should work until connections are evicted by the pool and then regenerated with the new password.

As for the event, I thought this will be more convenient because users might have their own logic/events for password change and this might be simpler and more straightforward.

dstepanov:
Can we test this with two datasource and validate that rerouting works?
Nevermind, I see we only support password change

jdbc-ucp/src/main/java/io/micronaut/configuration/jdbc/ucp/DatasourceFactory.java
```diff
@@ -39,12 +43,12 @@
...
  */
 @Factory
-public class DatasourceFactory implements AutoCloseable {
+public class DatasourceFactory implements AutoCloseable, ApplicationEventListener<DataSourcePasswordChangedEvent> {
     private static final Logger LOG = LoggerFactory.getLogger(DatasourceFactory.class);
     private final ApplicationContext applicationContext;
     private final UniversalConnectionPoolManagerConfiguration configuration;
 
-    private List<PoolDataSource> dataSources = new ArrayList<>(2);
+    private Map<String, PoolDataSource> dataSources = new HashMap<>(2);
```
dstepanov: Use Linked hash map

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceFactory.java
graemerocher: it isn't about being easier it is because this change won't work from the refresh management endpoint or wither periodic config refresh mechanisms

jdbc-ucp/src/main/java/io/micronaut/configuration/jdbc/ucp/DatasourceFactory.java
radovanradic: Changed

radovanradic:
> Can we test this with two datasource and validate that rerouting works? Nevermind, I see we only support password change

I have updates tests for DBCP, Hikari and Tomcat by altering default user in the test and then executing query. Something is not working as expected for UCP, investigating now.

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceFactory.java
radovanradic: Ok, I guess will wait this PR https://github.com/micronaut-projects/micronaut-core/pull/11798 to be merged and then for which drivers needed listen to refresh event and I suppose `datasources.*.password` and extract datasource name and update password if needed.

jdbc/src/main/java/io/micronaut/jdbc/BaseDatasourceFactory.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public void onApplicationEvent(RefreshEvent event) {
+        Map<String, Object> changes = event.getSource();
+        if (CollectionUtils.isEmpty(changes)) {
+            return;
+        }
+        for (String property : changes.keySet()) {
+            Matcher matcher = DATASOURCE_PASSWORD_MATCHER.matcher(property);
```
graemerocher: why only the password? One could change the username too no?

radovanradic: Yes, I suppose it can. Added in https://github.com/micronaut-projects/micronaut-sql/pull/1599/commits/7ff5f5dfeee644a54aee8b077c45618ade355113
Or that could be done in restart since probably won't happen frequently, but we can support it just could not add test like for password change.

jdbc/src/main/java/io/micronaut/jdbc/BaseDatasourceFactory.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * An abstract base class for datasource factories that listens for refresh events and updates datasource credentials accordingly.
+ * <p>
+ * This class provides a basic implementation for handling refresh events and updating datasource credentials.
+ * Subclasses are expected to implement the {@link #dataSourceCredentialsChanged(String, DataSourceCredentials)} method to handle the updated credentials.
+ */
+public abstract class BaseDatasourceFactory implements RefreshEventListener {
```
graemerocher, line 44: annotate with `@Internal` and javadoc `@since`

graemerocher (approved):

dstepanov (approved):

radovanradic:
Need to wait this PR https://github.com/micronaut-projects/micronaut-core/pull/11798 and micronaut-core version with it be propagated before can merge this

graemerocher:
Is this ready now?

radovanradic:
> Is this ready now?

UCP is behaving differently after upgrade to core 4.8.17 so I made some changes. However, currently native UCP tests failing and I am investigating how to fix it.

jdbc-ucp/src/main/java/io/micronaut/configuration/jdbc/ucp/DatasourceConfiguration.java
```diff
@@ -51,7 +52,7 @@
 public class DatasourceConfiguration implements BasicJdbcConfiguration {
     private static final Logger LOG = LoggerFactory.getLogger(DatasourceConfiguration.class);
 
-    @ConfigurationBuilder(allowZeroArgs = true, excludes = {"connectionFactoryProperties"})
+    @ConfigurationBuilder(allowZeroArgs = true, excludes = {"connectionFactoryProperties", "URL", "username", "password"})
```
radovanradic, line 55: @graemerocher This is what needed to change after upgrade to core 4.8.17 (containing fix for config refresh). So, when `bean.inject` is called from `RefreshScope` since this is `ConfigurationBuilder` it would call setter for all `PoolDataSourceImpl` bean and one of these are `setURL`, `setUsername`, `setPassword`. But, since we have these already in `DatasourceConfiguration` and are calling this delegate methods I think we don't need these.
Btw, when changed password in the database and refresh event is fired, calling `delegate.setURL` would stop and start connection pool but password has not been propagated yet. So, excluding these resolves the issue.
And in `DatasourceConfiguration.setUrl()` and other methods has now check to not call delegate if property has not changed since that would also trigger connection pool stop/start.

radovanradic:
Should be ready now, tests are passing

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceFactory.java
radovanradic: Changed logic in refresh event listener to get new value like 
`String password = applicationContext.getRequiredProperty(property, String.class);`
where property will be `datasources.{datasourcename}.password`

PR #1615, opened 2025-06-16
Workaround for issues with MSSQL and hibernate reactive

radovanradic:
In netty 4.2.2 SslContextBuilder defaults endpointIdentificationAlgorithm to "HTTPS" if system property "io.netty.handler.ssl.defaultEndpointVerificationAlgorithm" is not set. This makes `sun.security.util.HostnameChecker.match(String expectedName, X509Certificate cert, boolean chainsToPublicCA)` method to throw error 
>javax.net.ssl.SSLHandshakeException: No name matching localhost found

and this PR contains workaround that can be used as an example if users face similar issues.


radovanradic:
Not working for native tests :/
> Caused by: javax.net.ssl.SSLHandshakeException: No name matching localhost found

radovanradic:
> Not working for native tests :/
> 
> > Caused by: javax.net.ssl.SSLHandshakeException: No name matching localhost found

Fixed in https://github.com/micronaut-projects/micronaut-sql/pull/1615/commits/466a96a1876d19425865360b3477510acc321129

yawkat (approved):

Issue #1642, opened 2025-07-23
Unable to use @Replaces when the replacement extends the original factory.


henryju:
### Expected Behavior

I would like to replace the default datasource factory by a custom one. It seems to have been [possible](https://houseofclouds.se/increase-observability-for-your-application-using-aws-x-ray/) in the past with old Micronaut versions, but this is not working anymore:

```java
import com.amazonaws.xray.sql.TracingDataSource;
import io.micronaut.configuration.jdbc.hikari.DatasourceConfiguration;
import io.micronaut.configuration.jdbc.hikari.DatasourceFactory;
import io.micronaut.context.ApplicationContext;
import io.micronaut.context.annotation.Context;
import io.micronaut.context.annotation.EachBean;
import io.micronaut.context.annotation.Factory;
import io.micronaut.context.annotation.Replaces;
import javax.sql.DataSource;

@Factory
@Replaces(factory = DatasourceFactory.class)
public class XrayDataSourceFactory extends DatasourceFactory {
  public XrayDataSourceFactory(ApplicationContext applicationContext) {
    super(applicationContext);
  }

  @Override
  @Context
  @EachBean(DatasourceConfiguration.class)
  public DataSource dataSource(DatasourceConfiguration datasourceConfiguration) {
    return new TracingDataSource(super.dataSource(datasourceConfiguration));
  }
}
```

### Actual Behaviour

The original factory is still invoked, and the custom one ignored. This issue has been reported here: https://github.com/micronaut-projects/micronaut-core/issues/4054 but the issue has been closed. I commented there but in case it is missed I am reopening a new issue.

Using delegation is a possible workaround in my specific case, but not clean due to methods visibility issues. It also took me a lot of time to understand why it was not working.

### Version

4.9.1

radovanradic:
Likely not core issue but could be caused in recent PR in sql https://github.com/micronaut-projects/micronaut-sql/pull/1599

henryju:
I didn't try with older Micronaut versions, but in issue https://github.com/micronaut-projects/micronaut-core/issues/4054 OP said that the issue started when migrating from Micronaut 1.3.6 to 2.0.1. But maybe this is a different cause.

Do you want me to move the issue to micronaut-sql ?

radovanradic:
I have moved it to sql, but investigating now and maybe not caused by recent changes.

radovanradic:
It happens with Micronaut 3 as well 

[...]

PR #1643, opened 2025-07-24
Custom DatasourceFactory example and test

radovanradic:
Reproducer for the issue reported [here](https://github.com/micronaut-projects/micronaut-sql/issues/1642).


radovanradic:
First created this PR as reproducer for #1642
Turns out, to replace `DatasourceFactory` besides `@Replaces(DatasourceFactory.class)` the method creating `DataSource` needs `@Replaces(bean = DataSource.class, factory = DatasourceFactory.class)` so something like this 
```
@Context
    @EachBean(DatasourceConfiguration.class)
    @Replaces(bean = DataSource.class, factory = DatasourceFactory.class)
    public DataSource dataSource(DatasourceConfiguration datasourceConfiguration) {
        LOG.info("Created custom datasource");
        return new CustomHikariUrlDataSource(datasourceConfiguration);
    }
```

We can leave this as a test for custom data source.

graemerocher (approved):

PR #1645, opened 2025-07-25
add jooq-bom to version catalog

### LINKED ISSUE #1613: Use jooq-bom
### Feature description

Jooq released a new bom module
sprint-boot already started using it: https://github.com/spring-projects/spring-boot/commit/707cd9413663c9a392388bfa06ab0f4e305a445f

Even tho this project only uses `org.jooq:jooq` its nice having the same version as the bom so that you could get the bom with the same version as microanut without overriding it.

The current version doesn't have a bom, but this rennovate pr: https://github.com/micronaut-projects/micronaut-sql/pull/1589) does.

KDanisme:
the current jooq version doesn't have a bom so it is upgraded to the closest version that does.

fixes #1613

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -23,7 +23,7 @@ micronaut-validation = "4.9.0"
 # Frameworks
 
 managed-vertx = "4.5.16"
-managed-jooq = "3.19.18"
+managed-jooq = "3.19.24"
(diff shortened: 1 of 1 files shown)


sdelamo:
Thanks @KDanisme Closed via https://github.com/micronaut-projects/micronaut-sql/pull/1712

PR #1648, opened 2025-07-29
feat: Add Global Datasource configuration application for data-source-properties

jamesforwardnwboxed:
Right now this PR allows globally setting values in `data-source-properties`, via `global.datasources.data-source-properties`. We find this to be extremely helpful when enforcing organisational standards - for example, providing helm charts which provide these values out of the box which force connection pinning on PostgreSQL via `assumeMinServerVersion` and forcing the `ApplicationName` to match the provisioned microservice.

Naturally we could improve this in future to also allow global configuration of other properties rather than just the `data-source-properties` map, but this is a nice start

### DIFF
--- a/jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
+++ b/jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
@@ -199,7 +199,7 @@ public class DatasourceConfiguration extends BasicDataSource implements BasicJdb
         }
     }
 
-    public Map<String, ?> getIndividualDsProperties() {
+    Map<String, ?> getIndividualDsProperties() {
         return individualDsProperties;
     }
 }
--- a/jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
+++ b/jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
@@ -31,6 +31,8 @@ import org.slf4j.LoggerFactory;
 
 import jakarta.annotation.PostConstruct;
 import jakarta.annotation.PreDestroy;
+
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -52,6 +54,7 @@ public class DatasourceConfiguration extends BasicDataSource implements BasicJdb
     private static final Logger LOG = LoggerFactory.getLogger(DatasourceConfiguration.class);
     private final CalculatedSettings calculatedSettings;
     private final String name;
+    private Map<String, ?> individualDsProperties = new HashMap<>();
 
     /**
      * Constructor.
@@ -165,6 +168,7 @@ public class DatasourceConfiguration extends BasicDataSource implements BasicJdb
 
     @Override
     public void setDataSourceProperties(@MapFormat(transformation = MapFormat.MapTransformation.FLAT, keyFormat = StringConvention.RAW)  Map<String, ?> dsProperties) {
+        this.individualDsProperties = dsProperties;
         if (dsProperties != null) {
             dsProperties.forEach((s, o) -> {
                 if (o != null) {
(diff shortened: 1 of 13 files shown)


jamesforwardnwboxed:
Hi @graemerocher , I was wondering if it would be possible to get this merged in at some point? not sure who else to ping to get a review

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/GlobalDatasourceProperties.java
```diff
@@ -0,0 +1,56 @@
...
+ * repetition when multiple datasources share common settings such as
+ * connection pool parameters, SSL settings, or application-specific properties.</p>
+ *
+ * <p>This bean is only created when the "global.datasources.data-source-properties"
+ * configuration property is present, ensuring it doesn't interfere with applications
+ * that don't use global datasource configuration.</p>
+ *
+ * @author James Forward
+ */
+@Requires(property = "global.datasources.data-source-properties")
```
dstepanov, line 44: I think better name would be `global-data-source-properties`

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/GlobalDatasourceConfigModifier.java
```diff
@@ -0,0 +1,60 @@
...
+ * override global defaults. Properties with null values are ignored.</p>
+ *
+ * <p>This bean is only created when the "global.datasources.data-source-properties"
+ * configuration property is present.</p>
+ *
+ * @author James Forward
+ */
+@Requires(property = "global.datasources.data-source-properties")
+@Singleton
+public class GlobalDatasourceConfigModifier implements BeanCreatedEventListener<DatasourceConfiguration> {
```
dstepanov, line 42: Pls make final, package private and annotate `@Internal`

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/GlobalDatasourceProperties.java
```diff
@@ -0,0 +1,56 @@
...
+@ConfigurationProperties("global.datasources")
+public class GlobalDatasourceProperties {
```
dstepanov, line 46: Pls make final and add javadoc to methods

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/GlobalDatasourceProperties.java
```diff
@@ -0,0 +1,56 @@
...
+    private Map<String, String> dataSourceProperties = new HashMap<>();
```
dstepanov, line 47: Why this has to be mutatable?

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
```diff
@@ -194,4 +198,8 @@ void setEnabled(boolean enabled) {
             throw new DisabledBeanException("The datasource \"" + name + "\" is disabled");
         }
     }
+
+    Map<String, ?> getIndividualDsProperties() {
```
dstepanov, line 202: Why is this exposed?

dstepanov (changes requested):

dstepanov:
Instead of the listener you should inject the global conf and apply the settings in the post construct method.

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/GlobalDatasourceProperties.java
jamesforwardnwboxed: @dstepanov The idea was to have `global` at a higher level and then have the usual `datasources` properties underneath, which makes it extensible and more inline with the current setup, since these data source properties are already set at `datasources.<datasourcename>.data-source-properties`. For example, if we did want to set other properties that would usually be under `datasources.<datasourcename>`, such as a global username, it would be `global.datasources.username`

jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/DatasourceConfiguration.java
jamesforwardnwboxed: @dstepanov The dbcp implementation needs to access the properties to work properly

PR #1701, opened 2025-10-13
Update to micronaut core 4.10.4 

radovanradic:
Added workaround for processConfigProps task dealing with duplicate docs

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,7 +2,7 @@
 
 # Micronaut
 
-micronaut = "4.10.3"
+micronaut = "4.10.4"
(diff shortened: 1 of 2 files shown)


gradle/libs.versions.toml
```diff
@@ -2,7 +2,7 @@
 
 # Micronaut
 
-micronaut = "4.10.3"
+micronaut = "4.10.4"
 micronaut-docs = "2.0.0"
```
radovanradic, line 6: We could wait for docs 3.0.0 but probably not needed with the workaround for processConfigProps

graemerocher (approved):

PR #1722, opened 2025-11-05
Upgrade to Hibernate 7

radovanradic:
And Hibernate Reactive 4


gradle/libs.versions.toml
```diff
@@ -22,10 +22,10 @@ micronaut-validation = "4.11.0"
 
 # Frameworks
 
-managed-vertx = "4.5.21"
+managed-vertx = "5.0.5"
```
radovanradic, line 25: There is PR to upgrade vertx but is needed for Hibernate Reactive so can do it here.

hibernate-jpa/build.gradle
```diff
@@ -8,8 +8,12 @@ dependencies {
     api(libs.managed.hibernate.core)
     api(libs.managed.jakarta.transaction.api)
     api(mnValidation.micronaut.validation)
-    compileOnly(mnData.micronaut.data.tx.hibernate)
-    testImplementation(mnData.micronaut.data.tx.hibernate)
+    compileOnly(mnData.micronaut.data.tx.hibernate) {
+        exclude(group: "org.hibernate.orm")
```
radovanradic, line 12: I think it makes sense to exclude hibernate deps to transition from micronaut-data since this is main place where hibernate deps are declared.

buildSrc/src/main/groovy/io.micronaut.build.internal.test-application-hibernate.gradle
```diff
@@ -0,0 +1,8 @@
+plugins {
```
radovanradic, line 1: Instead having hibernate5 or hibernate6 in prev version, I think it makes sense have just hibernate since we support only current version of hibernate (this applies to tests etc)

hibernate-jpa/src/main/java/io/micronaut/configuration/hibernate/jpa/conf/serviceregistry/builder/supplier/internal/DefaultStandardServiceRegistryBuilderCreatorCreator.java
```diff
@@ -55,9 +52,6 @@ public DefaultStandardServiceRegistryBuilderCreatorCreator(@Primary @Nullable In
 
     @Override
     public StandardServiceRegistryBuilder create(JpaConfiguration jpaConfiguration) {
-        if (jpaConfiguration.isCompileTimeHibernateProxies()) {
```
radovanradic, line 58: org.hibernate.cfg.AvailableSettings.BYTECODE_PROVIDER is deprecated and will be removed
At present only bytebuddy is supported, bytebuddy being the default since version 5.3.
Hibernate ORM will use the BytecodeProvider implementation it finds on the
classpath loading it via the standard ServiceLoader mechanism. Currently, there is only a single
implementation which is included in Hibernate ORM, so it's not possible to override this.

dstepanov: It means no more:
```
jpa:
  default:
    compile-time-hibernate-proxies: true
```

radovanradic: It is still needed for this 
```
@Internal
@Prototype
final class CompileProxyConfigurer implements StandardServiceRegistryBuilderConfigurer {

    private final BeanProvider<BytecodeProvider> bytecodeProviderBeanProvider;

    CompileProxyConfigurer(BeanProvider<BytecodeProvider> bytecodeProviderBeanProvider) {
        this.bytecodeProviderBeanProvider = bytecodeProviderBeanProvider;
    }

    @Override
    public void configure(JpaConfiguration jpaConfiguration, StandardServiceRegistryBuilder standardServiceRegistryBuilder) {
        if (jpaConfiguration.isCompileTimeHibernateProxies()) {
            // It would be enough to add `ProxyFactoryFactory` by providing `BytecodeProvider` we eliminate bytecode Enhancer
            standardServiceRegistryBuilder.addInitiator(new StandardServiceInitiator<BytecodeProvider>() {
                @Override
                public BytecodeProvider initiateService(Map configurationValues, ServiceRegistryImplementor registry) {
                    return bytecodeProviderBeanProvider.get();
                }

                @Override
                public Class<BytecodeProvider> getServiceInitiated() {
                    return BytecodeProvider.class;
                }
            });
        }
    }

}
```
just this call 
```
System.setProperty(BYTECODE_PROVIDER, BYTECODE_PROVIDER_NAME_NONE);
```
does not seem to be doing anything and these references will be removed from Hibernate

dstepanov (approved):

dstepanov (approved):

PR #1746, opened 2025-11-10
Enable graalvm 25 

radovanradic:
I thought we might enable JDK 25 build for graalvm native to verify builds are passing for 25 as well.


buildSrc/src/main/groovy/io.micronaut.build.internal.test-application.gradle
```diff
@@ -37,6 +37,11 @@ graalvmNative {
     binaries {
         all {
             buildArgs.add("--report-unsupported-elements-at-runtime")
+            if (System.getenv("GRAALVM_HOME")?.contains("graal") && System.getenv("GRAALVM_HOME")?.contains("25")) {
+                logger.quiet("Enabling SharedArenaSupport for GraalVM 25")
+                buildArgs.add("-H:+SharedArenaSupport")
```
radovanradic: Need to enable this flag or else hibernate reactive builds are failing with the error

> Error: Exception in thread "vert.x-eventloop-thread-4" com.oracle.svm.core.jdk.UnsupportedFeatureError: Support for Arena.ofShared is not active: enable with -H:+SharedArenaSupport

graemerocher:
I feel like a lot of these changes should be in the project template / build tools. @melix WDYT?

melix:
We don't have any specific support for native in the build plugins, since these are on test projects. That said I'm not sure if the changes here should be done everywhere.

radovanradic:
Yes, I agree. This is needed only for hibernate reactive tests so I will try to move it there just where needed.

radovanradic:
graalvm 25 failing with latest jooq, requires bunch of build time initialization changes - resolved

radovanradic:
ping @graemerocher , @dstepanov

graemerocher (approved):

Issue #1779, opened 2025-12-11
Add support for unpooled JDBC datasources


rrobetti:
### Feature description

Problem
As the founder of Open J Proxy (OJP) (https://github.com/Open-J-Proxy/ojp), I need the ability to disable Micronaut's built-in connection pooling. OJP is a database proxy that provides virtual connections and handles the actual connection pooling at the proxy layer, outside of the application. When using OJP, having Micronaut manage its own connection pool creates unnecessary overhead and conflicts with OJP's connection management.

Currently, Micronaut SQL requires one of four connection pool implementations (HikariCP, Tomcat JDBC, Apache DBCP, or Oracle UCP). There is no way to create unpooled datasources where each getConnection() call creates a new physical connection and each close() actually closes it.

Comparison with Other Frameworks
Of the three major frameworks that OJP supports:

✅ Spring Boot: Provides DriverManagerDataSource for unpooled connections
✅ Quarkus: Supports unpooled datasources via Agroal configuration
❌ Micronaut: No option to disable connection pooling
Micronaut is currently the only framework without a way to turn off connection pooling, making it incompatible with OJP and similar database proxy solutions.

Use Cases
Beyond database proxy solutions like OJP, unpooled datasources are needed for:

Database Proxies: When using CNAM, pgpool, or Open J Proxy that manage connection pooling at the proxy layer
Testing: Integration tests that require clean connection state between tests
Serverless/FaaS: Short-lived functions where connection pool overhead outweighs benefits, especially when paired with database proxies
Very low-volume applications: Applications with < 1 request/second where pooling overhead is unnecessary
Educational/prototype applications: Learning JDBC without pool complexity
Proposed Solution
Implement a new micronaut-jdbc-unpooled module that provides a simple DataSource implementation using JDBC DriverManager. This would:

Create a new physical database connection for each getConnection() call
Actually close the connection when close() is called (not return to pool)
Follow existing Micronaut SQL module patterns (@EachProperty, BasicJdbcConfiguration, etc.)
Support all standard Micronaut features (credential refresh, multiple datasources, etc.)
Work seamlessly with Micronaut Data, Hibernate, jOOQ, and JDBI
Configuration
Usage would be identical to pooled datasources - just swap the dependency:

datasources:
  default:
    url: jdbc:postgresql://localhost:5432/mydb
    username: user
    password: pass
// Instead of:
// runtimeOnly("io.micronaut.sql:micronaut-jdbc-hikari")

// Use:
[...]

PR #1780, opened 2025-12-11
Add support for unpooled JDBC datasources

rrobetti:
Implements Issue: https://github.com/micronaut-projects/micronaut-sql/issues/1779


jdbc/src/main/java/io/micronaut/jdbc/UnpooledDatasourceFactory.java
```diff
@@ -0,0 +1,117 @@
...
+     * Creates an unpooled datasource for each {@link DatasourceConfiguration}.
+     *
+     * @param datasourceConfiguration The datasource configuration
+     * @return An unpooled {@link DataSource}
+     */
+    @Context
+    @EachBean(DatasourceConfiguration.class)
+    @Requires(condition = JdbcDataSourceEnabled.class)
+    public DataSource dataSource(DatasourceConfiguration datasourceConfiguration) {
+        UnpooledDataSource ds = new UnpooledDataSource(
```
dstepanov, line 72: Why do we need a delegating `UnpooledDataSource`? You should simply create a DataSource that is specified (maybe also add some builder option)

dstepanov (changes requested):

dstepanov:
Please sign CLA

jdbc/src/main/java/io/micronaut/jdbc/UnpooledDatasourceFactory.java
rrobetti: The rationale for implementing unpolled datasource support is documented here https://github.com/micronaut-projects/micronaut-sql/issues/1779.

dstepanov: Not sure I understand, please describe the reason

dstepanov:
Maybe we can implement it without an extra dependency, simply by enable it with some property WDYT @radovanradic

radovanradic:
> Maybe we can implement it without an extra dependency, simply by enable it with some property WDYT @radovanradic

Maybe adding `DatasourceFactory` in `jdbc` module that will also have `@EachBean(BasicJdbcConfiguration.class)` but with condition that `BasicJdbcConfiguration.disable-pool=true` ?

rrobetti:
> > Maybe we can implement it without an extra dependency, simply by enable it with some property WDYT @radovanradic
> 
> Maybe adding `DatasourceFactory` in `jdbc` module that will also have `@EachBean(BasicJdbcConfiguration.class)` but with condition that `BasicJdbcConfiguration.disable-pool=true` ?

thanks for your comment, how would that look like? Can you provide more info? Should I change this implementation?

radovanradic:
> > > Maybe we can implement it without an extra dependency, simply by enable it with some property WDYT @radovanradic
> > 
> > 
> > Maybe adding `DatasourceFactory` in `jdbc` module that will also have `@EachBean(BasicJdbcConfiguration.class)` but with condition that `BasicJdbcConfiguration.disable-pool=true` ?
> 
> thanks for your comment, how would that look like? Can you provide more info? Should I change this implementation?

I think what Denis suggested is there is no need for new module for this. Maybe you can just have UnpooledDatasourceFactory in jdbc module and then just create DataSource or as you named it UnpooledDataSource. And maybe there should be flag to tell whether datasource will be pooled to activate that factory, using condition similar to `io.micronaut.jdbc.JdbcDataSourceEnabled` but Denis might clarify more what he suggested.
Btw, this looks nice and simple to me.

rrobetti:
> > > > Maybe we can implement it without an extra dependency, simply by enable it with some property WDYT @radovanradic
> > > 
> > > 
> > > Maybe adding `DatasourceFactory` in `jdbc` module that will also have `@EachBean(BasicJdbcConfiguration.class)` but with condition that `BasicJdbcConfiguration.disable-pool=true` ?
> > 
> > 
> > thanks for your comment, how would that look like? Can you provide more info? Should I change this implementation?
> 
> I think what Denis suggested is there is no need for new module for this. Maybe you can just have UnpooledDatasourceFactory in jdbc module and then just create DataSource or as you named it UnpooledDataSource. And maybe there should be flag to tell whether datasource will be pooled to activate that factory, using condition similar to `io.micronaut.jdbc.JdbcDataSourceEnabled` but Denis might clarify more what he suggested.

@dstepanov can you add your take on this?

dstepanov:
I don’t think we need a new module, maybe just an option like “allow-not-pulling-datasource=true” globally or per datasouce

graemerocher:
done in a separate PR, see https://github.com/micronaut-projects/micronaut-sql/issues/1779

PR #1812, opened 2026-01-20
Set Oracle client name from micronaut.application.name property

### LINKED ISSUE #1446: Set Oracle client name from `micronaut.application.name` property
### Feature description

Micronaut has a global application name concept. Oracle connections can be given a session name that shows up in various places (e.g. when listing active sessions). It is easy to set:

```
datasources:
  default:
    url: jdbc:oracle:thin:@//localhost:1521/freepdb1
    driver-class-name: oracle.jdbc.OracleDriver
    username: abc
    password: abc
    dialect: oracle
    data-source-properties:
      v$session.program: "My Great Program"
```

.... but not _intuitive_ to set. Micronaut SQL should configure this data source property when the dialect is `oracle` by itself so everything just works.

radovanradic:
Implementation for #1446


graemerocher (approved):

mikehearn:
Is it possible to eliminate the code duplication?

radovanradic:
Cleaned up Sonar warnings

jdbc-hikari/src/main/java/io/micronaut/configuration/jdbc/hikari/DatasourceFactory.java
```diff
@@ -65,12 +66,40 @@ public DatasourceFactory(ApplicationContext applicationContext) {
...
+                LOG.debug("Skipping Oracle session program auto-config due to: {}", e.getMessage());
+            }
+        }
         HikariUrlDataSource ds = new HikariUrlDataSource(datasourceConfiguration);
         addMeterRegistry(ds);
         dataSources.put(datasourceConfiguration.getName(), ds);
         return ds;
     }
 
+    private void applyOracleSessionProgram(DatasourceConfiguration configuration) {
```
mikehearn: Could this method be moved to a utility class as it seems to be the same in each module?

mikehearn (changes requested):

jdbc-hikari/src/main/java/io/micronaut/configuration/jdbc/hikari/DatasourceFactory.java
radovanradic: It is not really the same implementation in all modules, can maybe improve some reusability.

mikehearn (approved):

PR #1813, opened 2026-01-20
fix: use DataSourceResolver instead of type check

abrenk:
Using instanceof does not work if the DataSource is wrapped by e.g. ContextualAwareDataSource or OpenTelemetryDataSource (or even multiple times).

I've added an implementation of `BeanCreatedListener<DataSource>` that just wraps the `DataSource` in a `TransactionAwareDataSourceProxy` as an example. And a new test class `ConnectionPoolManagerListenerSpec` that would fail if `instanceof` is used instead of `unwrap`.

### DIFF
--- a/jdbc-ucp/src/test/groovy/io/micronaut/configuration/jdbc/ucp/test/DataSourceWrapperBeanEventListener.java
+++ b/jdbc-ucp/src/test/groovy/io/micronaut/configuration/jdbc/ucp/test/DataSourceWrapperBeanEventListener.java
@@ -1,0 +1,30 @@
+package io.micronaut.configuration.jdbc.ucp.test;
+
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.context.event.BeanCreatedEvent;
+import io.micronaut.context.event.BeanCreatedEventListener;
+import io.micronaut.core.order.Ordered;
+import jakarta.inject.Singleton;
+import org.springframework.jdbc.datasource.TransactionAwareDataSourceProxy;
+
+import javax.sql.DataSource;
+
+/**
+ * A {@link BeanCreatedEventListener} that wraps each {@link DataSource} in a {@link TransactionAwareDataSourceProxy}
+ * to create an example of a wrapped {@link DataSource} for test cases.
+ */
+@Singleton
+@Requires(env = "test")
+public class DataSourceWrapperBeanEventListener implements BeanCreatedEventListener<DataSource>, Ordered {
+
+    public DataSource onCreated(BeanCreatedEvent<DataSource> event) {
+        final DataSource dataSource = event.getBean();
+
+        return new TransactionAwareDataSourceProxy(dataSource);
+    }
+
+    public int getOrder() {
+        return Ordered.HIGHEST_PRECEDENCE;
+    }
+
+}
--- a/jdbc-ucp/src/main/java/io/micronaut/configuration/jdbc/ucp/ConnectionPoolManagerListener.java
+++ b/jdbc-ucp/src/main/java/io/micronaut/configuration/jdbc/ucp/ConnectionPoolManagerListener.java
@@ -20,8 +20,10 @@ import io.micronaut.context.event.BeanCreatedEvent;
 import io.micronaut.context.event.BeanCreatedEventListener;
 import io.micronaut.context.exceptions.ConfigurationException;
 import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.order.Ordered;
 import io.micronaut.core.util.StringUtils;
+import io.micronaut.jdbc.DataSourceResolver;
 import jakarta.inject.Singleton;
 import oracle.ucp.UniversalConnectionPoolAdapter;
 import oracle.ucp.UniversalConnectionPoolException;
@@ -31,6 +33,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.sql.DataSource;
+import java.sql.SQLException;
 
 
 /**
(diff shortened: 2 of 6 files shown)


graemerocher:
it might be better to use https://github.com/micronaut-projects/micronaut-sql/blob/13fb2553de270dab1d2a0c13c68487f9813b690e/jdbc/src/main/java/io/micronaut/jdbc/DataSourceResolver.java#L26

abrenk:
I'm sorry, I don't understand... you mean it would be better to use an injected `DataSourceResolver` and its `DataSourceResolver.resolve` method instead of `DataSource.unwrap` (or `instanceof`)? I fear, I don't really see what `DataSourceResolver` is to be used for and what implementation will be available. The empty DEFAULT instance does nothing and `DelegatingDataSourceResolver` (the only implementation I can find) wouldn't help either with unwrapping a `PoolDataSource` from a `ContextualAwareDataSource`, `TransactionAwareDataSourceProxy` or `OpenTelemetryDataSource` or their cascades. `DataSource.unwrap` really is the only way, I think.

radovanradic:
You could thy injecting `List<DataSourceResolver> dataSourceResolversList<DataSourceResolver> dataSourceResolvers` and then unwrap like this 
```
for (DataSourceResolver dataSourceResolver : dataSourceResolvers) {
            dataSource = dataSourceResolver.resolve(dataSource);
        }
```

abrenk:
I've taken a deep look at this and now understand more of the `DataSourceResolver` interface and how it is used in e.g. the micronaut-jdbc-dbcp and micronaut-jdbc-tomcat modules.

But although the `DelegatingDataSourceResolver` implementation present in the ApplicationContext of the application I've debugged handles the `ContextualAwareDataSource` (as it extends `DelegatingDataSource`), it does *not* handle the `OpenTelemetryDataSource` (that is present when OTel is enabled, see [`DataSourceBeanCreatedEventListener`](https://github.com/micronaut-projects/micronaut-tracing/blob/v7.2.0/tracing-opentelemetry-jdbc/src/main/java/io/micronaut/tracing/opentelemetry/instrument/jdbc/DataSourceBeanCreatedEventListener.java)) and the resolve loop stops there.

That's why `ConnectionPoolManagerListener` would not see the `PoolDataSource` after calling `resolve` but still the `OpenTelemetryDataSource` and therefor would still not work as expected.

What is the correct solution to this, @graemerocher?

- Adding a `DataSourceResolver` like in other modules, calling its `resolve` method and _then_ `DataSource.unwrap(PoolDataSource.class)`? This is the only solution I can think of if you want me to add the resolve call.

- Or does the OTel integration in micronaut-tracing need to provide a `DataSourceResolver` that handles `OpenTelemetryDataSource`? How does an ApplicationContext with multiple implementations work? No code I've seen in Micronaut handles this like @radovanradic proposed above. As soon as I add a second `DataSourceResolver` I get a `NonUniqueBeanException`:
```
Message: Multiple possible bean candidates found: [OpenTelemetryDataSourceResolver, DelegatingDataSourceResolver]
Path Taken: 
new @j.i.Singleton i.m.m.h.m.HealthMonitorTask(CurrentHealthStatus currentHealthStatus, List<HealthIndicator> healthIndicators)
\---> new @j.i.Singleton i.m.m.h.m.HealthMonitorTask(CurrentHealthStatus currentHealthStatus, [List<HealthIndicator> healthIndicators])
      \---> new @j.i.Singleton i.m.m.h.i.j.JdbcIndicator(@Named("blocking") ExecutorService executorService, DataSource[] dataSources, [DataSourceResolver dataSourceResolver], HealthAggregator<HealthResult> healthAggregator)
...
```

abrenk:
Brainstorming some ideas:

- `io.micronaut.tracing.opentelemetry.instrument.jdbc.DataSourceBeanCreatedEventListener` could return an implementation of `DelegatingDataSource` that delegates all `DataSource` method calls to `OpenTelemetryDataSource` but implements `getTargetDataSource` to return the original instance. But `DelegatingDataSource` is a class and not an interface and its `getTargetDataSource` method is `final`, so it's somewhat annoying to implement without more changes.
- An implementation of `DataSourceResolver` that `@Replaces` the `DelegatingDataSourceResolver` to extend it for `OpenTelemetryDataSource`. But that does not scale for further cases.

radovanradic:
If you inject list of DataSourceResolver then there would not be multiple bean issue and you can iterate through the list and unwrap?

abrenk:
> If you inject list of DataSourceResolver then there would not be multiple bean issue and you can iterate through the list and unwrap?

But then _all_ other usages of `DataSourceResolver` (e.g. the `JdbcIndicator` the exception cited above was for) would need to be changed, too.

abrenk:
I pushed an implementation using `resolve` + `unwrap` as a fallback.

In order to remove the `unwrap` fallback the framework needs a `CompositeDataSourceResolver` like `CompositeDiscoveryClient` or `CompositeHttpLocaleResolver`. I will provide an implementation in this pull request but can split it off into a separate pull request if you prefer.

In a pull request for [micronaut-tracing](https://github.com/micronaut-projects/micronaut-tracing) I will add a `OpenTelemetryDataSourceResolver` implementation that pulls the delegate from `OpenTelemetryDataSource`. Since there's no accessor for that field I'll have to use reflection like in https://github.com/micronaut-projects/micronaut-sql/blob/7.0.x/jdbc-dbcp/src/main/java/io/micronaut/configuration/jdbc/dbcp/metadata/DbcpDataSourcePoolMetadata.java#L93.

That should be the best solution and match the existing architecture.

abrenk:
I have now implemented a `CompositeDataSourceResolver` and added it to the PR. Similar to `CompositeHttpLocaleResolver`, it iterates over all available `DataSourceResolver` beans. It returns the first `DataSource` that is not the same as the method's parameter.

This solves the issue I described earlier: it allows the resolution to continue even when multiple proxies (like OpenTelemetry or ContextualAwareDataSource) are wrapping the actual PoolDataSource. _And while writing this I see another problem - the composite implementation needs to loop until all resolvers return the same object!_ I'll fix that.

With this in place, the UCP integration can rely entirely on the `DataSourceResolver` interface without needing direct unwrap calls or worrying about proxy-layering. (As soon as micronaut-tracing adds an `OpenTelemetryDataSourceResolver`.) This should make the implementation robust and idiomatic.

abrenk:
`CompositeDataSourceResolver` now continues resolving until no further unwrapping is possible.

jdbc-ucp/src/test/groovy/io/micronaut/configuration/jdbc/ucp/test/DataSourceWrapperBeanEventListener.java
```diff
@@ -0,0 +1,30 @@
...
+import javax.sql.DataSource;
+
+/**
+ * A {@link BeanCreatedEventListener} that wraps each {@link DataSource} in a {@link TransactionAwareDataSourceProxy}
+ * to create an example of a wrapped {@link DataSource} for test cases.
+ */
+@Singleton
+@Requires(env = "test")
+public class DataSourceWrapperBeanEventListener implements BeanCreatedEventListener<DataSource>, Ordered {
+
```
radovanradic, line 19: The test is failing because there is no resolver to unwrap this, so might need to add something like this 
```
@Singleton
class TransactionAwareDataSourceProxyResolver implements DataSourceResolver {

    @Override
    public DataSource resolve(DataSource dataSource) {
        if (dataSource instanceof TransactionAwareDataSourceProxy transactionAwareDataSourceProxy) {
            return transactionAwareDataSourceProxy.getTargetDataSource();
        }
        return dataSource;
    }
}
```

radovanradic:
I think this looks good. We have another `@Primary` resolver in micronaut-micrometer https://github.com/micronaut-projects/micronaut-micrometer/blob/6.0.x/micrometer-observation-datasource/src/main/java/io/micronaut/micrometer/observation/datasource/ProxyDataSourceResolver.java which will need to be changed after this. It needed to run all resolvers before this one because it checks instance so I guess after unwrapping with new `CompositeDataSourceResolver` and declaring this resolver in micrometer as lowest order would work the same way.

abrenk:
I had exactly the same code for `TransactionAwareDataSourceProxyResolver` locally but forgot to add it to the commit. 🤦‍♂️

Now it's present. I've rebased onto the current 7.0.x branch and changed the order of the commits to introduce the `CompositeDataSourceResolver` first.

You might want to change the title of this pull request to "fix: use DataSourceResolver instead of type check".

radovanradic (approved):

graemerocher:
Thanks for the contribution!

PR #1913, opened 2026-04-15
Add regression for executor transaction isolation

### LINKED ISSUE #1508: Transaction can't spin off additional, unrelated transactional work via ExecutorService
### Expected Behavior

In Micronaut 3, I could write code that would execute in a transactional context and then submit work via an `ExecutorService` to happen in it's own, separate transaction. Our application makes heavy use of this pattern and it seems reasonable to expect it to work in Micronaut 4.

Instead, in Micronaut 4, it appears that because of `ExecutorServiceInstrumenter` the context from the initial transaction is dragged into the downstream runnable. The result is when any database work starts (including the attempt to start a new transaction), it fails with an error from Hikari that "the connection is closed".

### Actual Behaviour

Two separate transactions are able to execute.

### Steps To Reproduce

Run the test in https://github.com/lightbody/micronaut-data-4-tx-issue

### Environment Information

_No response_

### Example Application

https://github.com/lightbody/micronaut-data-4-tx-issue

### Version

4.x

graemerocher:
## Summary
Adds JDBI transaction coverage for executor-submitted work started from `afterCommit`, asserting the follow-up write runs in an independent transaction.

## Verification
- `./gradlew :micronaut-jdbi:txTest --console=plain`

Resolves #1508

### DIFF
--- a/jdbi/src/txTest/groovy/io/micronaut/configuration/jdbi/example/ApplicationSpec.groovy
+++ b/jdbi/src/txTest/groovy/io/micronaut/configuration/jdbi/example/ApplicationSpec.groovy
@@ -1,5 +1,6 @@
 package io.micronaut.configuration.jdbi.example
 
+import io.micronaut.configuration.jdbi.example.jdbitransaction.ExecutorTransactionIsolationService
 import io.micronaut.configuration.jdbi.example.jdbitransaction.ConcurrentTransactionsBug
 import io.micronaut.context.ApplicationContext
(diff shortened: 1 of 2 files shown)


graemerocher:
@copilot apply changes based on the comments in [this thread](https://github.com/micronaut-projects/micronaut-sql/pull/1913#pullrequestreview-4115086776)

radovanradic (approved):

PR #1929, opened 2026-04-20
Fix JPA configuration when there is no primary/default JPA configuration in 6.3.x

radovanradic:
Apply fix from 7.x https://github.com/micronaut-projects/micronaut-sql/pull/1811

### DIFF
--- a/buildSrc/src/main/groovy/io.micronaut.build.internal.sql-module.gradle
+++ b/buildSrc/src/main/groovy/io.micronaut.build.internal.sql-module.gradle
@@ -5,6 +5,7 @@ plugins {
 
 ossIndexAudit {
     excludeCoordinates = [
+            "org.hibernate.reactive:hibernate-reactive-core:2.4.17.Final", // no 2.14.x version patched https://ossindex.sonatype.org/component/pkg:maven/org.hibernate.reactive/hibernate-reactive-core
             "com.h2database:h2:2.3.232" // no version patched https://ossindex.sonatype.org/component/pkg:maven/com.h2database/h2
     ]
 }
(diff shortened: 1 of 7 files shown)


radovanradic:
Seems like we cannot create patch as hibernate-reactive has vulnerabilities (even latest 2.4.17 that we could use here) and GraalVM runners fail with the error for jdk 17
> Error: Feature defined by io.micronaut.core.io.service.ServiceLoaderFeature unexpectedly failed with a(n) java.lang.UnsupportedClassVersionError. Please report this problem to the authors of io.micronaut.core.io.service.ServiceLoaderFeature.

sdelamo (approved):

sdelamo:
@radovanradic 

> Error: Feature defined by io.micronaut.core.io.service.ServiceLoaderFeature unexpectedly failed with a(n) java.lang.UnsupportedClassVersionError. Please report this problem to the authors of io.micronaut.core.io.service.ServiceLoaderFeature.

I set [target and source compatibility](https://github.com/micronaut-projects/micronaut-sql/pull/1929/commits/4f1fe1fd08e8a0dca285bd138bafd76d85284e22) and it seems to workaround that issue.

sdelamo:
@radovanradic 
> Seems like we cannot create patch as hibernate-reactive has vulnerabilities (even latest 2.4.17 that we could use here) 

There is indeed no patched 2.4.x version of hibernate-reactive. [I accepted the version](https://github.com/micronaut-projects/micronaut-sql/pull/1929/commits/32e0756762276e65b9b60db3ec89a255aed1fa5e).

PR #2086, opened 2026-08-21
Add MyBatis native metadata and improve native-image test support

radovanradic:
Added MyBatis tests in 7.1.0 did not have nativeTest task. Figured out when tried to run guides with micronaut platform 5.1.1 and it has duplicate MyBatis feature. When used MyBatis feature from the starter then it failed on native tests. This adds metadata and changes MyBatis java test to native. Created PR in shared metadata repo as well: https://github.com/oracle/graalvm-reachability-metadata/pull/9477


mybatis/src/main/java/io/micronaut/configuration/mybatis/MyBatisMapperScan.java
```diff
@@ -0,0 +1,53 @@
...
+ * Requests compile-time discovery of MyBatis mapper interfaces in the given packages.
+ *
+ * @since 7.2.0
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Introduction
+@Singleton
+public @interface MyBatisMapperScan {
```
radovanradic, line 37: Needed this since configuring by package was not working on graalvm native.

mybatis-processor/src/main/java/io/micronaut/configuration/mybatis/processor/MyBatisMapperScanVisitor.java
```diff
@@ -0,0 +1,133 @@
...
+ */
+public final class MyBatisMapperScanVisitor implements TypeElementVisitor<Object, Object> {
+
+    private final Set<String> mapperTypes = new LinkedHashSet<>();
+    private final Map<String, Scan> scans = new LinkedHashMap<>();
+    private boolean processed;
+
+    @Override
+    public VisitorKind getVisitorKind() {
+        return VisitorKind.AGGREGATING;
```
graemerocher, line 56: this doesn't aggregate anything right now since it doesn't write anything out that aggregates what it is collecting. It only mutates metadata.

To make this mean something you need to write an aggregating result

apart from that getSupportedAnnotationNames() needs to be overridden and implemented

src/main/docs/guide/mybatis/mybatis-customizers.adoc
```diff
@@ -1,12 +1,38 @@
...
+.Groovy
+----
+compileOnly("io.micronaut.sql:micronaut-mybatis-processor")
+----
+
+[source,kotlin]
+.Kotlin
+----
+ksp("io.micronaut.sql:micronaut-mybatis-processor")
+----
```
graemerocher, line 34: I believe the dependency:: macro with an annotationProcessor scope already writes out the right thing so you don't have to duplicate these snippets

graemerocher (changes requested):

src/main/docs/guide/mybatis/mybatis-customizers.adoc
radovanradic: An annotationProcessor scope renders `annotationProcessor(...)`; it does not map to `compileOnly` for Groovy or ksp for Kotlin. Since this processor requires different configurations for each language, the language-specific snippets are needed.

mybatis-processor/src/main/java/io/micronaut/configuration/mybatis/processor/MyBatisMapperScanVisitor.java
radovanradic: Implemented `getSupportedAnnotationNames()` and changed the aggregating visitor to generate a mapper-registration result instead of only mutating annotation metadata.

mybatis-processor/src/main/java/io/micronaut/configuration/mybatis/processor/MyBatisMapperScanVisitor.java
```diff
@@ -0,0 +1,337 @@
...
+            throw new ProcessingException(originatingElement, "Failed to generate '" + classDef.getName() + "': " + e.getMessage(), e);
+        }
+    }
+
+    /**
+     * MyBatis implements mapper interfaces with {@link java.lang.reflect.Proxy} and instantiates and populates
+     * result and parameter objects reflectively. Both need GraalVM metadata, which is written next to the
+     * generated registration so that users do not have to declare it by hand.
+     */
+    private static void writeNativeImageMetadata(VisitorContext context, NativeImageMetadata metadata) {
```
graemerocher: I think this part is problematic, since it might conflict with what GraalReflectionConfigurer writes. We should likely upstream into core a way to register proxied types rather than potentially duplicating this logic here

radovanradic: Ok, switched to adding `@ReflectionConfig` values to the annotated type so `GraalTypeElementVisitor` generates the configurer; the processor writes no metadata files itself anymore. Needed a separate isolating visitor because the aggregating processor runs after the Graal visitor on javac.
Native tests are passing

